PR #4, opened 2016-08-18
use webpack with "library" options for build

jebeck:
OK, here it is @krystophv 

I created a separate config that just builds on the main one a little to keep it cleaner for making changes in the main config as far as loaders. We *do* use the main config for React Storybook.

The main thing that's confusing me a bit is that two copies of React problem is still an issue. I would have that defining React as an `external` would have solved that, but I guess when webpack resolves the dependency, it still looks in the "closest" `node_modules` dir? Not an issue since you may have found another solution to that, just adds to my lack of confidence that I got things right here...

In terms of merge order, this should be merged along with [tideline #283](https://github.com/tidepool-org/tideline/pull/283) and [blip #318](https://github.com/tidepool-org/blip/pull/318), but in terms of testing it out locally, you should try it with branches that actually _use_ the viz repo, which means [tideline jebeck/cgm-trends-iteration.2](https://github.com/tidepool-org/tideline/pull/286), [blip jebeck/cgm-trends-iteration.2](https://github.com/tidepool-org/blip/pull/320) and here [jebeck/cgm-trends-RC](https://github.com/tidepool-org/viz/tree/jebeck/cgm-trends-RC).

Hope this make sense, LMK if there are questions...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.DS_Store
 coverage/
 dist/
 node_modules/
--- a/package.config.js
+++ b/package.config.js
@@ -1,0 +1,25 @@
+const _ = require('lodash');
+const baseConfig = require('./webpack.config');
+
+const packageConfig = _.cloneDeep(baseConfig);
+
+packageConfig.output.libraryTarget = 'commonjs';
+
+packageConfig.externals = {
+  classnames: 'classnames',
+  'd3-array': 'd3-array',
+  'd3-format': 'd3-format',
+  'd3-scale': 'd3-scale',
+  'd3-shape': 'd3-shape',
+  'd3-time': 'd3-time',
+  lodash: 'lodash',
+  'moment-timezone': 'moment-timezone',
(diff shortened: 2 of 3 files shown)


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;qQqZiU3z&#x2F;3-ux-2-dev-8-tst-3-blip-cgm-in-trends-mvp-release-candidate"> &lt;ux:2 dev:8 tst:3&gt; [blip] [cgm in trends] MVP release candidate</a>

jebeck:
FYI, I still expect the build on jebeck/cgm-trends-iteration.2 in blip to be failing (like it is) because I haven't published a new version of the viz package to npm. But I'm trying to get blip jebeck/prep-for-viz to pass - since it doesn't actually depend on viz yet, it should. Just playing whack-a-minor-mistake at the moment I think.

jh-bate:
LGTM

PR #7, opened 2016-08-23
deps updates

jebeck:
Since we just did another pass on blip & tideline. This matches those updates (plus I updated the things that are only here, like Storybook).

Ping @jh-bate or @krystophv for review. We should merge this in and then updates the branches we're working on.


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;zhbSxi4k&#x2F;27-viz-update-deps-to-match-latest-round-of-blip-tideline-deps-updates"> [viz] update deps to match latest round of blip + tideline deps updates</a>

jh-bate:
LGTM

PR #8, opened 2016-08-25
Pump settings - WIP

jh-bate:
@krystophv just getting this up in the hope you can start a first pass review


src/containers/settings/factory.js
```diff
@@ -0,0 +1,35 @@
...
+  if (chartType === 'carelink') {
+    return Medtronic;
+  } else if (chartType === 'tandem') {
+    return Tandem;
+  } else if (chartType === 'insulet') {
+    return Omnipod;
+  } else if (chartType === 'animas') {
+    return Animas;
+  }
+  return null;
```
jebeck: Should we perhaps throw an error in the `else` condition? I think having an error modal pop up would probably be a better experience than just a blank display on the settings if this ever happened in production...

src/containers/settings/data/utilities.js
```diff
@@ -0,0 +1,103 @@
...
+
+export const DISPLAY_PRESCION_PLACES = 3;
+export const MGDL_UNITS = 'mg/dL';
+export const MMOLL_UNITS = 'mmol/L';
+
+function noData(val) {
+  return val === null || (typeof val === 'undefined');
+}
+
+export function getFormattedTime(startTime) {
```
jebeck: I'm not a fan of this approach adding another layer of wrapping a function that's just in another directory of the same repo. Over time, this will make it harder to read components and understand which functions are being used across multiple components and may very easily lead to duplicate code. In the trends work, I've been [importing the specific functions I need](https://github.com/tidepool-org/viz/blob/jebeck/cgm-trends-iteration.2/src/components/trends/cbg/FocusedCBGSliceHTMLLabels.js#L146) in that component from the datetime utilities.

src/containers/settings/data/utilities.js
```diff
@@ -0,0 +1,103 @@
...
+    schedules.push({ name: names[i], position: i });
+  }
+  return schedules;
+}
+
+export function getDeviceMeta(settingsData) {
+  return {
+    name: settingsData.deviceId || 'unknown',
+    schedule: settingsData.activeSchedule || 'unknown',
+    uploaded: datetime.formatDisplayDate(settingsData.deviceTime) || 'unknown',
```
jebeck: We don't *ever* use `deviceTime` for display - this is the raw device display timestamp and could be very inaccurate if the user had one of those big mistakes like setting their device to the wrong year.

What should be done is to translate the `time` (this is the bootstrapped value) into a local time using the user's current timezone from their `timePrefs`, which get passed down from blip. There is already a function `getTimezoneFromTimePrefs` in datetime for getting the user's timezone from the `timePrefs` (it's more complex than you might think since there's a `timezoneAware` preference as well). Then you'll need to adjust the `formatDisplayDate` function in datetime to include setting the timezone: `moment.utc(settings.time).tz(timezone).format('MMM Do YYYY')`.

src/containers/settings/data/utilities.js
```diff
@@ -0,0 +1,103 @@
...
+ */
+
+import _ from 'lodash';
+
+import * as datetime from '../../../utils/datetime';
+import * as format from '../../../utils/format';
+
+export const DISPLAY_PRESCION_PLACES = 3;
+export const MGDL_UNITS = 'mg/dL';
+export const MMOLL_UNITS = 'mmol/L';
```
jebeck: Move these constants to a general file, perhaps src/utils/constants.js? They will be useful in far more places than just the settings work (in fact, I can immediately refactor to use them in the trends work).

src/containers/settings/data/utilities.js
```diff
@@ -0,0 +1,103 @@
...
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+import _ from 'lodash';
+
+import * as datetime from '../../../utils/datetime';
+import * as format from '../../../utils/format';
+
+export const DISPLAY_PRESCION_PLACES = 3;
```
jebeck: Sp. on 'precision'.

src/components/common/Table.js
```diff
@@ -0,0 +1,101 @@
...
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ *
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+import React, { PropTypes } from 'react';
+
+class Table extends React.Component {
```
jebeck: I don't see a reason why this can't be a stateless functional component... Am I missing something?

I think the only reason the linter isn't complaining is because of the instance methods, but that feels a bit like cheating. Thoughts?

src/containers/settings/header/Header.js
```diff
@@ -0,0 +1,65 @@
...
+class Header extends React.Component {
+
+  constructor(props) {
+    super(props);
+    this.handleClick = this.handleClick.bind(this);
+    this.state = { serialNumberExpanded: false };
+  }
+
+  handleClick() {
+    this.setState({ serialNumberExpanded: !this.state.serialNumberExpanded });
```
jebeck: So we should clarify requirements for this. If there isn't an expectation that the expansion to show SN is preserved when you navigate around the different views (e.g., to daily view then back), then handling it with local state like this is fine. If, OTOH, there **is** such an expectation of state preservation, then we should do this via redux. We should get the same clarity around expectations of state preservation of collapsed vs. expanded sections for the Tandem timed settings "profiles" and the different basal profiles for the other pumps. That'll be a larger task to redux-ify, so maybe as a later task to be prioritized.

src/containers/settings/animas/Animas.js
```diff
@@ -0,0 +1,171 @@
...
+  const { bgUnits, pumpSettings } = props;
+
+  const renderBasalsData = () => {
+    const columns = [
+      { key: 'start', label: 'Start time', className: '' },
+      { key: 'rate', label: 'Value (U/hr)', className: '' },
+    ];
+    const schedules = utilities.getScheduleNames(pumpSettings.basalSchedules);
+
+    const tables = schedules.map((schedule) => {
```
jebeck: I would have preferred picking a standard here between native `.map` and `_.map`, but once again we've gone different directions. How important do we think consistency is here? It trips me up sometimes because of the different arguments (callback alone vs. callback second).

src/components/common/Table.js
```diff
@@ -0,0 +1,101 @@
...
+      </table>
+    );
+  }
+}
+
+Table.propTypes = {
+  title: PropTypes.object,
+  rows: PropTypes.array,
+  columns: PropTypes.array,
+  tableStyle: PropTypes.string,
```
jebeck: Shouldn't the props be required? Even if the component can render w/o error without any props, it's useless to render nothing, so I tend to think we should require everything that's necessary for the *expected* display under a "normal" use case. Remember that props validation doesn't do anything in prod (and doesn't *throw* errors in any case, just logs them), so it won't cause errors for the user even if something goes awry.

src/containers/common/CollapsibleContainer.js
```diff
@@ -0,0 +1,80 @@
...
+    this.handleClick = this.handleClick.bind(this);
+  }
+
+  handleClick() {
+    const current = this.state.isOpened;
+    this.setState({ isOpened: !current });
+  }
+
+  render() {
+    let label = (<div className="label" onClick={this.handleClick}>{this.props.label}</div>);
```
jebeck: Why the non-CSS modules `className` here?

src/containers/settings/data/processing.js
```diff
@@ -0,0 +1,222 @@
...
+  basalSchedules: React.PropTypes.arrayOf(
+    React.PropTypes.shape({
+      name: React.PropTypes.string.isRequired,
+      value: React.PropTypes.arrayOf(
+        React.PropTypes.shape({
+          start: React.PropTypes.number.isRequired,
+          rate: React.PropTypes.number.isRequired,
+        }),
+      ),
+    }),
```
jebeck: I think you need to add an `isRequired` on this line to make sure the array contains at least one object with the given shape.

src/containers/settings/data/processing.js
```diff
@@ -0,0 +1,222 @@
...
+    amount: utilities.getBloodGlucoseValue(
+      sensitivityData,
+      'amount',
+      startTime,
+      bgUnits
+    ),
+  }));
+}
+
+export function processTimedSettings(pumpSettings, schedule, bgUnits) {
```
jebeck: So since this one is totally specific to Tandem, I personally would probably just keep it in the Tandem.js component. I tend to feel only shared things belong factored out. Thoughts?

src/index.js
```diff
@@ -24,16 +24,22 @@ import TwoOptionToggle from './components/common/controls/TwoOptionToggle';
...
+
+import CollapsibleContainer from './containers/common/CollapsibleContainer';
+
 import vizReducer from './reducers/';
 
 const components = {
   FocusedCBGSliceHTMLLabels,
   FocusedCBGSliceTime,
   TwoOptionToggle,
+  CollapsibleContainer,
```
jebeck: Why is this exported? It's not needed in blip so far as I can tell...

src/utils/datetime.js
```diff
@@ -68,3 +68,17 @@ export function formatDurationToClocktime(duration) {
...
+    return '';
+  }
+  return moment(timestamp).utc().format('MMM Do YYYY');
+}
+
+export function millisecondsAsTimeOfDay(milliseconds) {
+  if (milliseconds === null || milliseconds === undefined) {
+    return '';
+  }
+  return moment(milliseconds).utc().format('hh:mm a');
```
jebeck: You have duplicated here. My `formatDurationToClocktime` serves the same purpose, although my use case was a bit more complex so it provides the result a little differently. Let's discuss how to consolidate for all use cases.

src/utils/format.js
```diff
@@ -17,9 +17,16 @@
 
 import { format } from 'd3-format';
 
+export function displayDecimal(val, places) {
```
jebeck: Not a huge fan of this function name/overloading. Confusing to use a function called `displayDecimal` for displaying an integer when no `places` provided. Can we at least think of a different name?

jebeck:
I focused comments on repo organization/standards stuff and looking at the utilities. Have not looked at tests or in detail at components, so if you haven't yet @krystophv, those would be good for you to look at.

krystophv:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;tbc9evMS&#x2F;6-ux-1-dev-5-tst-1-blip-device-settings-update-design-for-tandem-pump-device-settings"> &lt;ux: 1 dev:5 tst: 1&gt; [blip] [device settings] Update design for Tandem pump Device Settings</a>

src/containers/settings/data/utilities.js
jh-bate: agreed - have pushed those into `constants.js`

src/containers/settings/data/utilities.js
jh-bate: I have removed this now

src/containers/settings/data/utilities.js
jh-bate: its only needed here so no longer exported and fixed the spelling

src/containers/settings/header/Header.js
jh-bate: ta

src/containers/settings/animas/Animas.js
jh-bate: @krystophv did you research which is the most efficient? Happy switch but will wait until we have decided which way to go

src/containers/settings/data/processing.js
jh-bate: moved as per props discussion already

src/components/common/Table.js
jh-bate: I have update to require the types we need

src/index.js
jh-bate: removed now

src/utils/settings/data.js
```diff
@@ -0,0 +1,348 @@
...
+ * @param  {Object} settingsData all settings data
+ * @param  {Object} timePrefs    timezone preferences object
+ *
+ * @return {Object}              filtered meta data
+ */
+export function getDeviceMeta(settingsData, timePrefs) {
+  let uploadedTime;
+  if (timePrefs.timezoneAware) {
+    if (settingsData.time) {
+      uploadedTime = datetime.formatDisplayDate(moment(settingsData.time).valueOf(), timePrefs);
```
jebeck: There's no utility to using moment here and it may slow things down - moment has had egregious problems with slow parsing of timestamps in the past (and I don't know if they've been fixed). Only doing it once here, but for consistency let's only use moment where absolutely necessary == when doing timezone-related manipulations. The best alternative here for `moment(settingsData.time).valueOf()` is just `Date.parse(settingsData.time)`.

src/utils/settings/data.js
```diff
@@ -0,0 +1,348 @@
...
+    } else {
+      uploadedTime = 'unknown';
+    }
+  } else {
+    if (settingsData.deviceTime) {
+      uploadedTime = datetime.formatDisplayDate(moment(settingsData.deviceTime).valueOf(),
```
jebeck: Ditto ;)

src/utils/datetime.js
```diff
@@ -110,3 +110,18 @@ export function millisecondsAsTimeOfDay(milliseconds, format = 'h:mm a') {
...
+/**
+ * formatDisplayDate
+ * @param  {Number} timestamp integer hammertime
+ * @param  {Object} timePrefs object containing timezoneAware Boolean and timezoneName String
+ *
+ * @return {String}           formatted timezoneAware date string
+ */
+export function formatDisplayDate(timestamp, timePrefs) {
+  if (!timestamp) {
+    return '';
```
jebeck: Can you look at the other functions for a model of how to handle incorrect inputs rather than returning an empty string? @jh-bate and I looked at this when combining the two functions from the branches that had the (now) `millisecondsAsTimeOfDay` functionality. My experience with this tricky time stuff is we don't want these functions to silently fail; let's be consistent about throwing informative errors if the inputs are not as expected. If you look at a method like `localNoonBeforeTimestamp` we're being specified about *not* getting a`Date` (because of the historical issues with the `Date` constructor you and I discussed yesterday that have made me eternally wary). I think it'd be great to align (and I can update code too) re: what else to error on, like complete lack of input and whether we'll even accept a string rather than an integer hammertime.

src/utils/datetime.js
```diff
@@ -110,3 +110,18 @@ export function millisecondsAsTimeOfDay(milliseconds, format = 'h:mm a') {
...
+  }
+  return moment(timestamp).utc().tz(getTimezoneFromTimePrefs(timePrefs))
```
jebeck: Let's get to the bottom of this issue with passing arguments to `moment` rather than `utc` and be consistent about it. Happy to help investigate.

test/utils/settings/data.test.js
```diff
@@ -0,0 +1,183 @@
...
+    it('should return the name, schedule and date uploaded device', () => {
+      expect(
+        data.getDeviceMeta(settingsData, timePrefs)
+      ).to.have.property('name').equal('DevId0987654321');
+      expect(
+        data.getDeviceMeta(settingsData, timePrefs)
+      ).to.have.property('schedule').equal('Normal');
+      expect(
+        data.getDeviceMeta(settingsData, timePrefs)
+      ).to.have.property('uploaded').not.equal('unknown');
```
jebeck: Can you please make this more specific than just "not unknown"?

test/utils/datetime.test.js
```diff
@@ -210,4 +209,39 @@ describe('datetime', () => {
...
+    };
+    const tzAwareNY = {
+      timezoneAware: true,
+      timezoneName: 'America/New_York',
+    };
+    const tzUnaware = {
+      timezoneAware: false,
+      timezoneName: null,
+    };
+    const hammertime = 1473048000000; // 09/05/2016 @ 4:00am (UTC)
```
jebeck: Would you mind reformatting the comment in ISO 8601 to avoid confusing people like @jh-bate who aren't used to seeing month first?

src/utils/settings/data.js
```diff
@@ -0,0 +1,348 @@
...
+        timePrefs);
+    } else {
+      uploadedTime = 'unknown';
+    }
+  }
+  return {
+    name: settingsData.deviceId || 'unknown',
+    schedule: settingsData.activeSchedule || 'unknown',
+    uploaded: uploadedTime,
```
jebeck: For consistency, could we just delay the assignment of 'unknown' for `uploadedTime` to here?

test/utils/settings/data.test.js
```diff
@@ -0,0 +1,183 @@
...
+  });
+  describe('getDeviceMeta', () => {
+    const timePrefs = {
+      timezoneAware: false,
+      timezoneName: null,
+    };
+    it('should return the name, schedule and date uploaded device', () => {
+      expect(
+        data.getDeviceMeta(settingsData, timePrefs)
+      ).to.have.property('name').equal('DevId0987654321');
```
jebeck: This test should be failing because you shouldn't be displaying the `deviceId` as the serial number. Looks like @krystophv did the work to make sure serial number is available but it hasn't been surfaced in the rendering.

src/utils/datetime.js
krystophv: It's really strange. I put together a set of console logs to try and figure out what it was doing, but it's clear that the `moment().utc(<arg>)` version is significantly different from the `moment(<arg>).utc()` version. Maybe you can make better sense of what's happening, but it seems to be somehow relative to the current time if we do `moment().utc(<arg>)` since the two `.valueOf()` calls are one millisecond apart, inferring some execution time passed between them.
```
time (string):                 2016-09-13T16:59:52.000Z
hrmt - Date.parse(time):       1473785992000
moment().utc(time).valueOf():  1473880342561
moment().utc(hrmt).valueOf():  1473880342562
moment(time).utc().valueOf():  1473785992000
moment(hmrt).utc().valueOf():  1473785992000
moment(time).format():         2016-09-13T12:59:52-04:00
moment(hmrt).format():         2016-09-13T12:59:52-04:00
moment(time).utc().format():   2016-09-13T16:59:52Z
moment(hmrt).utc().format():   2016-09-13T16:59:52Z
moment().utc(time).format():   2016-09-14T19:12:22Z
moment().utc(hmrt).format():   2016-09-14T19:12:22Z
moment().utc().format():       2016-09-14T23:12:22Z
```

jebeck: What is `hrmt`? I can't see where you've defined it.

krystophv: hmrt is defined in the second line - abbr. of hammertime, result of Date.parse(time)

test/utils/settings/data.test.js
krystophv: Actually that's fine, because `name` correlates to `deviceId`. We need to add some tests for the `serial` property.

jebeck: But you won't need `deviceId` anymore, right? Can't this just be removed from the `getDeviceMeta` method?

krystophv: I couldn't find any more references to the `.name` property, so I dropped it.

src/utils/settings/data.js
krystophv: Yup, done.

test/utils/datetime.test.js
krystophv: Since the function now accepts either hammertime or a zulu timestamp string, we now set this from the ISO 8601 string which resides directly above it for clarity.

test/utils/settings/data.test.js
krystophv: Now with significantly more specificity.

src/utils/datetime.js
krystophv: Modeled changes after functions which accept either Zulu timestamp string or hammertime.

src/utils/settings/data.js
krystophv: Dropped it, made changes as recommended.

src/utils/settings/data.js
krystophv: Ditto

src/utils/datetime.js
jebeck: So the issue here is that `moment().utc()` is not the same as `moment.utc()` - what I've been using is the latter, and I think my reasons for doing so are sound. I've documented here as a starting point for discussion and (hopefully) convergence: http://momentjs.com/docs/#/parsing/utc/

krystophv: Yup, seems to be the issue. I've altered the call to be in line with the rest of the repo.

PR #9, opened 2016-09-07
ESLint additions

jebeck:
Strawman proposal for additions to our ESLint config.

Added the plugin for lodash to prefer *all* lodash methods over native. (This will be a lot of changes. We can be more nuanced if we want. Let's discuss.)

Also added an additional config in utils/ to add the rule (only in that dir) to require JSDoc comments.

CC @krystophv @jh-bate Please try it out with an experimental merge into your working branch and let's discuss.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,5 +4,8 @@
   "plugins": ["lodash"],
   "rules": {
     "lodash/prefer-lodash-method": "error"
+  },
+  "settings": {
+    "lodash": 3
   }
 }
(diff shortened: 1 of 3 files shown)


jebeck:
After some brief discussion @krystophv and I have decided I'll merge this to master and update my trends work, then perhaps @jh-bate can update the settings work later. There will be a fair few changes now, but they should still go relatively quickly. The payoff of an enforceable standard for the future should be worth it!

PR #10, opened 2016-09-14
add instructions for publishing to npm

jebeck:
Merge if you don't have questions/comments/additions @jh-bate 

Ping @krystophv to take a look, need to get your npm account (or you need to create one) so we can add you to the org.


jh-bate:
LGTM

PR #11, opened 2016-09-15
document utilities import/export standards and proposed use of moment

jebeck:
A starting point for discussion so we can converge on something consistent @krystophv @jh-bate 

Let me know your thoughts and perhaps let's discuss after scrum today?


src/utils/README.md
```diff
@@ -0,0 +1,32 @@
...
+### use of the `moment` API for datetime manipulation
+
+We use [`moment-timezone`](http://momentjs.com/timezone/) when we need to do datetime manipulation involving an arbitrary timezone since neither JavaScript `Date` or D3 provide this functionality. We also try to *only* use `moment` where necessary, preferring JavaScript `Date` or D3 where one or the other is sufficient.
+
+We are trying to move towards a standard in this repository of dealing with datetimes as hammertimes (milliseconds since January 1, 1970) as much as possible, rather than as ISO 8601-formatted Zulu timestamps in String format, as we've done in the past.
+
+When you need to apply an arbitrary timezone to format a datetime for display, the initial construction of the moment object should look like this, if `utc` is your hammertime value:
+
+```JavaScript
+const d = moment.utc(utc);
```
jh-bate: `uct` as `hamertime` or `hammertimeUtc` ?

src/utils/README.md
```diff
@@ -0,0 +1,32 @@
...
+```
+
+See [moment's documentation on 'UTC' mode](http://momentjs.com/docs/#/parsing/utc/) for further details.
+
+In particular, try to avoid `moment(utc).utc()`, as this will first create a representation of the hammertime in the browser's timezone/locale, then convert it back to offset-less UTC, an unnecessary step.
```
jh-bate: can we codify this check with https://github.com/wunderflats/eslint-plugin-moment-utc/blob/master/docs/rules/no-moment-without-utc.md ?

src/utils/README.md
krystophv: Is `hammertimeUtc` redundant since Unix time doesn't concern itself with timezones? I feel like maybe just 'hammertime' should be descriptive enough.

jebeck: How's this for a naming proposal:
- `Date.parse(d.time)` => `hammertime`
- `Date.parse(d.deviceTime + '.000Z')` => `rawDisplayTimeMs`

krystophv: Seems alright to me.

jebeck:
Updated README now that we've settled on `hammertime` and tried out the ESLint plugin - works like a charm!! So have added that as well (only put it in the `utils/` .eslintrc because we should be keeping our use of moment limited to that directory anyway, but LMK if you object to that @jh-bate and @krystophv 

Will merge shortly.

PR #12, opened 2016-09-15
render smbg range overlay

jebeck:
This is a first chunk of the work to port the smbg version of the trends view into this repo. Here we have the stats calculation for the smbg display and render the min/max overlay and mean circle with animated transition as you change day filters and/or navigate along the time dimension.

Ping @jh-bate for review and I can now show you a refactoring task that you could take to build off this.


src/containers/trends/SMBGRangeAvgAnimationContainer.js
```diff
@@ -0,0 +1,130 @@
...
+    smbgRangeOverlay: PropTypes.bool.isRequired,
+    tooltipLeftThreshold: PropTypes.number.isRequired,
+    unfocusRange: PropTypes.func.isRequired,
+    xScale: PropTypes.func.isRequired,
+    yScale: PropTypes.func.isRequired,
+  };
+
+  static defaultProps = {
+    // three-hour bins
+    binSize: 1000 * 60 * 60 * 3,
```
jh-bate: is it worth using our three hours const I spied somewhere?

jh-bate: this one `import { THREE_HRS } from '../../utils/datetime';`

test/containers/trends/SMBGRangeAvgAnimationContainer.test.js
```diff
@@ -0,0 +1,91 @@
...
+} = scales.trends;
+
+import SMBGRangeAvgAnimationContainer
+  from '../../../src/containers/trends/SMBGRangeAvgAnimationContainer';
+
+describe('SMBGRangeAvgAnimationContainer', () => {
+  let wrapper;
+
+  // six-hour bins for testing
+  const binSize = 1000 * 60 * 60 * 6;
```
jh-bate: just noticed else where you where doing `THREE_HRS * 6` should we just do the same here as it is a multiple of 3?

test/containers/trends/SMBGRangeAvgAnimationContainer.test.js
```diff
@@ -0,0 +1,91 @@
...
+        .to.deep.equal({ mungedData: [] });
+    });
+  });
+
+  describe('componentWillReceiveProps', () => {
+    it('remunges data if binSize has changed', () => {
+      const instance = wrapper.instance();
+      sinon.spy(instance, 'mungeData');
+      expect(instance.mungeData.callCount).to.equal(0);
+      wrapper.setProps({ binSize: 1000 * 60 * 60 * 3 });
```
jh-bate: now `THREE_HRS`

test/containers/trends/TrendsSVGContainer.test.js
```diff
@@ -45,9 +48,18 @@ describe('CBGTrendsContainer', () => {
       veryLowThreshold: 60,
     },
     bgUnits: 'mg/dL',
```
jh-bate: use existing const for consistency?

package.json
```diff
@@ -103,6 +103,7 @@
     "d3-scale": "1.0.3",
     "d3-shape": "1.0.3",
     "d3-time": "1.0.2",
+    "react-dimensions": "2.0.0-alpha1",
```
jebeck: I think we're fine since we only support Chrome for the dataviz so any issues should easily come up during QA, although perhaps I'll make a note to test in Chrome on Windows as well. I've also looked at the repo's GitHub issues, and nothing huge stands out. v2 has breaking changes and is basically a better approach, and for our use case has produced no problems so far.

src/containers/trends/SMBGRangeAvgAnimationContainer.js
jebeck: Done.

src/containers/trends/TrendsContainer.js
```diff
@@ -305,40 +305,24 @@ export class TrendsContainer extends React.Component {
...
-    }
-    return null;
+    return (
+      <TrendsSVGContainer
+        bgBounds={this.props.bgBounds}
+        bgUnits={this.props.bgUnits}
+        data={this.state.currentCbgData}
+        focusedSlice={this.props.viz.trends.focusedCbgSlice}
+        focusedSliceKeys={this.props.viz.trends.focusedCbgSliceKeys}
+        focusRange={(d) => { console.log('focusRange', d); }}
```
jebeck: Yep.

src/containers/trends/TrendsSVGContainer.js
```diff
@@ -0,0 +1,220 @@
...
+};
+
+TrendsSVGContainer.propTypes = {
+  bgBounds: PropTypes.shape({
+    veryHighThreshold: PropTypes.number.isRequired,
+    targetUpperBound: PropTypes.number.isRequired,
+    targetLowerBound: PropTypes.number.isRequired,
+    veryLowThreshold: PropTypes.number.isRequired,
+  }).isRequired,
+  bgUnits: PropTypes.oneOf(['mg/dL', 'mmol/L']).isRequired,
```
jebeck: No, this is just a new component that I'd started before we merged together our work. I've fixed.

test/containers/trends/SMBGRangeAvgAnimationContainer.test.js
jebeck: From here below all your comments are about constants in tests. We discussed this a bit with the units constants before. I would have a lot of updating to do to use constants in the tests - I tend to feel it is overkill in this case (and makes test setups take more time), although I won't stop you from using them if you prefer.

Happy to be outvoted on this one for consistency's sake. Thoughts, @krystophv?

krystophv: Generally I feel like the tests should mimic the code path of the actual execution as closely as possible. If the expected use of the code makes direct use of a constant, then we should probably also use that constant in the test. If it's just a derived test fixture that _could_ make use of a constant, I'm pretty ¯\\\_(ツ)\_/¯ about it.

PR #13, opened 2016-09-16
rework trends state reducer

jebeck:
This is a fairly substantial reworking of how the state is stored for a particular "view" - in this case trends view.

The initial implementation was very naive and wasn't even clearing on logout(!), but this version stores all trends-view related state as a hash map by userId (so we can persist it if a clinician navigates to a different patient then back to the first, for example). We also make sure to clear it when required (on logout) and not clear it when undesirable (on re-fetch of patient's data).

@jh-bate this should actually be your model for the redux work on the smbg trends stuff, at least if @krystophv agrees with the approach - would like his eyes and thoughts on this as well.


jebeck:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;VJ1JQDTe&#x2F;34-viz-bugfix-for-showing-cgm-trends-view-even-when-not-enough-data-after-logout-and-switching-patients"> [viz] bugfix for showing CGM trends view even when not enough data after logout and switching patients</a>

src/containers/trends/TrendsContainer.js
```diff
@@ -342,17 +341,24 @@ export class TrendsContainer extends React.Component {
...
 }
 
-export function mapDispatchToProps(dispatch) {
+export function mapDispatchToProps(dispatch, ownProps) {
   return bindActionCreators({
-    focusTrendsCbgSlice: actions.focusTrendsCbgSlice,
-    markTrendsViewed: actions.markTrendsViewed,
-    unfocusTrendsCbgSlice: actions.unfocusTrendsCbgSlice,
+    focusTrendsCbgSlice: actions.focusTrendsCbgSlice.bind(
+      null, ownProps.currentPatientInViewId
```
krystophv: Is the `null` here explicit/intentioned or is this a good place to use `_.partial` - https://lodash.com/docs/3.10.1#partial ?

jebeck: Yeah I suppose but we'd want to be consistent about it, no? Doesn't look like we can enforce a standard with [eslint-plugin-lodash](https://github.com/wix/eslint-plugin-lodash). Shall I try to change all instances (I know I do it elsewhere, probably in CBGSlice.js) and we'll all try to keep an eye out in reviews in future or should I leave them all?

src/redux/constants/actionTypes.js
```diff
@@ -18,3 +18,9 @@
 export const FOCUS_TRENDS_CBG_SLICE = 'FOCUS_TRENDS_CBG_SLICE';
 export const MARK_TRENDS_VIEWED = 'MARK_TRENDS_VIEWED';
 export const UNFOCUS_TRENDS_CBG_SLICE = 'UNFOCUS_TRENDS_CBG_SLICE';
+
+// from blip's redux implementation
+// TODO: how could we DRY this out??
```
jebeck: Yeah, I can't think of another way to handle "gracefully" other than a common package. I fear it may come up more with the viz-blip interface, but probably unlikely that it'll happen with more than a half dozen action types, so perhaps that'll all right, not a huge quantity of redundant code.

jebeck:
Did you have any thoughts on the quite large difference in approach here @krystophv? I was uncertain about this as I was doing it Friday because it feels like the reducers (and testing them) just got a lot more complex, but I can't/couldn't on first attempt really see a simpler way to achieve the results we want.

src/containers/trends/TrendsContainer.js
krystophv: I would prefer the `_.partial` over the `.bind(null, ...)` as they are actually doing slightly different things and I was sort of left wondering why we needed to drop the execution context from this set of binds.

src/redux/constants/actionTypes.js
krystophv: I think in this case, it'll be alright.

krystophv:
It doesn't seem unnecessarily complex to me. Much of the uploader's state is managed in a similar way, with a hashmap keyed by userId - it seems reasonable to me to do the same here.

krystophv:
LGTM 👍

PR #14, opened 2016-09-21
Smbg focusing range mean slice

jh-bate:
actions and reducers for SMBG range avg


jh-bate:
<a href="https:&#x2F;&#x2F;trello.com&#x2F;c&#x2F;35wkckjP&#x2F;1-viz-port-smbg-version-of-trends-view-to-viz-repo-pt-1"> [viz] port smbg version of trends view to viz repo, Pt. 1</a>

src/redux/reducers/trendsStateByUser.js
```diff
@@ -20,13 +20,15 @@ import update from 'react-addons-update';
 
 import * as actionTypes from '../constants/actionTypes';
 
-const FOCUSED_SLICE = 'focusedCbgSlice';
-const FOCUSED_KEYS = 'focusedCbgSliceKeys';
+const FOCUSED_CBG_SLICE = 'focusedCbgSlice';
+const FOCUSED_CBG_KEYS = 'focusedCbgSliceKeys';
+const FOCUSED_SMBG_AVG = 'focusedSmbgAvg';
```
jebeck: Can we do `FOCUSED_SMBG_RANGE_AVG` and `focusedSmbgRangeAvg` like elsewhere, for consistency?

src/redux/reducers/trendsStateByUser.js
```diff
@@ -47,8 +49,17 @@ const trendsStateByUser = (state = {}, action) => {
...
+          [FOCUSED_CBG_KEYS]: { $set: focusedKeys },
+        } }
+      );
+    }
+    case actionTypes.FOCUS_TRENDS_SMBG_RANGE_AVG: {
+      const { rangeAvgData: data, rangeAvgPosition: position, userId } = action.payload;
+      return update(
+        state,
+        { [userId]: {
+          [FOCUSED_SMBG_AVG]: { $set: { data, position } },
```
jebeck: I like that you've used the simple `data` for the name of the state tree field, but now I think where I used `slice` will be confusing. Can you go through and make the changes to make the cbg equivalent parallel - `data` instead of `slice`? (This will also mean updating the places where we _use_ this state from the tree, which is in both viz and blip. I can help you generate a list of places to modify if you don't want to just change it in the reducer and then play error whack-a-mole 😉)

src/redux/reducers/trendsStateByUser.js
jh-bate: done

src/redux/reducers/trendsStateByUser.js
jh-bate: yeah if you can give me a list quickly I will get that done

jebeck: OK, I think the components are:
- containers/trends/CBGSlicesAnimationContainer
- components/trends/cbg/FocusedCBGSliceTime
- component/trends/cbg/FocusedCBGSliceHTMLLabels

No blip code changes after all - the latter two components are *rendered* via blip but obviously the code for them is in viz.

darinkrauss:
[<0:ux 5:dev 2:tst> [platform] Delete account and all related data](https://trello.com/c/ER64SIAH/8-0-ux-5-dev-2-tst-platform-delete-account-and-all-related-data)

PR #17, opened 2016-09-23
ensure time binning is >= start and < end for each slice

jebeck:
What it says on the tin. (Before slicing was > start and <= end which is a bit unintuitive although it usually doesn't matter, but for Animas data with only minute resolution where you actually get data points on the border between slices, it matters.)

For your reviewing pleasure, @jh-bate or @krystophv

### DIFF
--- a/test/utils/trends/data.test.js
+++ b/test/utils/trends/data.test.js
@@ -49,12 +49,12 @@ describe('[trends] data utils', () => {
         expect(utils.findTimeOfDayBin(binSize, 0)).to.equal(1800000);
       });
 
-      it('should assign a bin of `1800000` to a datum at time 3600000', () => {
-        expect(utils.findTimeOfDayBin(binSize, 3600000)).to.equal(1800000);
+      it('should assign a bin of `1800000` to a datum at time 3599999', () => {
+        expect(utils.findTimeOfDayBin(binSize, 3599999)).to.equal(1800000);
       });
 
-      it('should assign a bin of `5400000` to a datum at time 3600001', () => {
-        expect(utils.findTimeOfDayBin(binSize, 3600001)).to.equal(5400000);
+      it('should assign a bin of `5400000` to a datum at time 3600000', () => {
+        expect(utils.findTimeOfDayBin(binSize, 3600000)).to.equal(5400000);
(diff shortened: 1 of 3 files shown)


jebeck:
[<ux:2 dev:8 tst:3> [blip] [cgm in trends] MVP release candidate](https://trello.com/c/qQqZiU3z/6-ux-2-dev-8-tst-3-blip-cgm-in-trends-mvp-release-candidate)

test/utils/trends/data.test.js
```diff
@@ -49,12 +49,12 @@ describe('[trends] data utils', () => {
         expect(utils.findTimeOfDayBin(binSize, 0)).to.equal(1800000);
       });
 
-      it('should assign a bin of `1800000` to a datum at time 3600000', () => {
-        expect(utils.findTimeOfDayBin(binSize, 3600000)).to.equal(1800000);
+      it('should assign a bin of `1800000` to a datum at time 3599999', () => {
+        expect(utils.findTimeOfDayBin(binSize, 3599999)).to.equal(1800000);
```
jh-bate: am I reading this wrong?
`findTimeOfDayBin(...)` 
- the size is one hour based one the test constant `binSize`
- in this test we are giving it  a time of `3599999`  
- based on the name of the function I expect to be returned a time
- but in the test we say the the returned value `1800000` is the bin

I know I am likely misreading it but thought I should say it now

jebeck: The bins are times - the time at the center of each slice. Also the function is meant to be read: find time-of-day bin (e.g., find bin for this particular time of day), not find time of day-bin or something like that. The return is supposed to be a bin, not a time. But the bins are times anyway :P

jh-bate: `findBinForTimeOfDay(...)`

jebeck: OK, I've made that change, but beware it will break things on every other branch we have in flight (where it's used for the smbg side of things in `SMBGRangeAvgAnimationContainer`, which I can't update from this branch which doesn't have that code yet).

jebeck:
@krystophv can you review for sign-off?

PR #19, opened 2016-09-26
Missing data message

jh-bate:
@jebeck


jh-bate:
[[viz] port smbg version of trends view to viz repo, Pt. 1](https://trello.com/c/35wkckjP/1-viz-port-smbg-version-of-trends-view-to-viz-repo-pt-1)

src/containers/trends/CBGSlicesAnimationContainer.js
```diff
@@ -141,13 +140,12 @@ export default class CBGSlicesAnimationContainer extends React.Component {
...
     if (_.isEmpty(mungedData)) {
-      const { margins, svgDimensions } = this.props;
-      const xPos = (svgDimensions.width / 2) - margins.left + margins.right;
-      const yPos = (svgDimensions.height / 2) - margins.top + margins.bottom;
       return (
-        <text className={styles.noDataMsg} id="noDataMsg" x={xPos + 40} y={yPos}>
-          No CGM data for this time period :(
-        </text>
+        <NoData
+          message="No CGM data for this time period :("
```
jebeck: I'm quite surprised you didn't factor this out as a constant! Also, the message is the same for both except for the CGM vs. BGM difference, so perhaps let's just make a template string with a gap for CGM or BGM a defaultProp in NoData and pass in the data type? (And have `cbg` map to 'CGM' and `smbg` map to 'BGM' in some kind of set of constants...)

test/components/trends/common/NoData.test.js
```diff
@@ -0,0 +1,88 @@
...
+    right: 0,
+    bottom: 0,
+    left: 0,
+  };
+  const dimensionsProp = {
+    width: 100,
+    height: 50,
+  };
+
+  it('should render without issue when all properties provided', () => {
```
jebeck: I haven't been doing these sorts of tests anymore with stubbing `console.error` - it's insanely problematic because React suppresses later calls to `console.error`. If you feel strongly that we should keep this test, at least add some other check that the component is rendered?

test/components/trends/common/NoData.test.js
```diff
@@ -0,0 +1,88 @@
...
+    shallow(
+      <NoData
+        message={messageProp}
+        margins={marginsProp}
+        dimensions={dimensionsProp}
+      />
+    );
+    expect(console.error.callCount).to.equal(0);
+  });
+  it('should render with one issue when no message provided', () => {
```
jebeck: Ditto. Try reordering this test with the one above and then trying to make the first one fail by not passing any props - the number of errors will not be what you expect.

jebeck: Also I think what I suggested above (having the message be part of a default prop) will remove the need for these kinds of tests anyway ;)

test/components/trends/common/NoData.test.js
jh-bate: yeah ok - that is probably something I adopted from blip

jebeck: Yup, we do it there. But it's a complete nightmare with test ordering.

CC @hntrdglss probably good to read this exchange ^ and understand we're trying to move away from this style of React rendering test.

src/components/trends/common/NoData.js
```diff
@@ -0,0 +1,50 @@
...
+
+  return (
+    <text className={styles.noDataMsg} id="noDataMsg" x={position.x} y={position.y}>
+      {noDataMessage({ type: dataType })}
+    </text>
+  );
+};
+
+NoData.defaultProps = {
+  dataType: 'CBG',
```
jebeck: I think you may have misunderstood one of my previous comments. I don't think the *datatype* should be a default prop, but rather the message template itself as `messageTemplate` or similar (so we could override it as/if needed). Perhaps another default prop `displayTypes` can contain the user-facing datatype for each of the `cbg` and `smbg` types (maybe `{cbg: 'CGM', smbg: 'BGM'}`), but let's just pass in `cbg` or `smbg` as a required prop `type` (or `datatype` if React throws an error when you tried to use `type` - I can't remember if that might be a problem).

Also when I run locally I'm getting the default "CBG" even for BGM, so the active type is not being passed down.

src/components/trends/common/NoData.js
```diff
@@ -0,0 +1,50 @@
...
+ * == BSD2 LICENSE ==
+ */
+
+import _ from 'lodash';
+import React, { PropTypes } from 'react';
+
+import styles from './NoData.css';
+
+const NoData = (props) => {
+  const noDataMessage = _.template('There is no <%= type %> data for this time period :(');
```
jebeck: You don't need underscore for this... Please use an ES6 template string: `There is no ${props.dataType} for this time period :(`

jebeck: Ah, my bad, you can't put a template string in a default prop. Keep the lodash then, but move the template function to a defaultProp plz.

src/containers/trends/TrendsSVGContainer.js
```diff
@@ -67,9 +68,27 @@ export class TrendsSVGContainer extends React.Component {
...
+        position={messagePosition}
+      />
+    );
+  }
+
   renderCbg() {
-    const { containerHeight: height, containerWidth: width } = this.props;
     if (this.props.showingCbg) {
+      if (_.isEmpty(this.props.cbgData)) {
+        return this.renderNoDataMessage();
```
jebeck: You're not passing the `dataType` argument for `renderNoDataMessage`, which is why I'm always getting the CBG display...

src/containers/trends/CBGSlicesAnimationContainer.js
```diff
@@ -141,13 +140,11 @@ export default class CBGSlicesAnimationContainer extends React.Component {
...
     const { mungedData } = this.state;
     if (_.isEmpty(mungedData)) {
-      const { margins, svgDimensions } = this.props;
-      const xPos = (svgDimensions.width / 2) - margins.left + margins.right;
-      const yPos = (svgDimensions.height / 2) - margins.top + margins.bottom;
       return (
-        <text className={styles.noDataMsg} id="noDataMsg" x={xPos + 40} y={yPos}>
-          No CGM data for this time period :(
-        </text>
+        <NoData
```
jebeck: Please remove the rendering of `NoData` from this component entirely. The task here was to move the rendering up a level so that it's being handled in common for both cbg and smbg versions of the trends view in `TrendsSVGContainer`.  It looks like that's been done, so cleanup here please :)

src/containers/trends/TrendsSVGContainer.js
jebeck: And related: I think it should be pretty easy to set up some tests with the right props, so that we can *test* that we're rendering the correct no data message for cbg vs. smbg. Since this is a current 🐛 in the code, I think this would be a good idea.

src/containers/trends/TrendsSVGContainer.js
```diff
@@ -67,9 +68,27 @@ export class TrendsSVGContainer extends React.Component {
     ]);
   }
 
+  renderNoDataMessage(dataType) {
+    const { containerHeight: height, containerWidth: width, margins } = this.props;
+    const xPos = (width / 2) - margins.left + margins.right;
+    const yPos = (height / 2) - margins.top + margins.bottom;
```
jebeck: Now that this component is being rendered one level higher in the tree, it needs to be shifted to the left and down because now it's *outside* the group where we already had the margins included. So basically just delete the subtraction of the margins here and it should be centered.

This is your code:
screenshot 2016-10-17 17 23 46

This is prod:
screenshot 2016-10-17 17 23 57

test/containers/trends/CBGSlicesAnimationContainer.test.js
```diff
@@ -81,7 +81,7 @@ describe('CBGSlicesAnimationContainer', () => {
       const instance = wrapper.instance();
       sinon.spy(instance, 'mungeData');
       expect(instance.mungeData.callCount).to.equal(0);
-      wrapper.setProps({ data: [{ id: 'a1b2c3', msPer24: 0, value: 90 }] });
+      wrapper.setProps({ data: [{ id: 'a2b3c4', msPer24: 6000, value: 180 }] });
```
jh-bate: I will have a look - maybe it was a merge thing as I didn't intentionally change it ??

src/containers/trends/CBGSlicesAnimationContainer.js
jh-bate: yup :)

test/containers/trends/CBGSlicesAnimationContainer.test.js
jebeck: Ah, that seems likely actually. No worries then.

jh-bate:
@jebeck updates are all there - use string message rather than function as it made more sense to keep it encapsulated to me

src/containers/trends/TrendsSVGContainer.js
```diff
@@ -41,13 +41,14 @@ const SMBG_OPTS = {
...
 import dimensions from 'react-dimensions';
-
+import _ from 'lodash';
 
 import { MGDL_UNITS, MMOLL_UNITS } from '../../utils/constants';
 import { THREE_HRS } from '../../utils/datetime';
 import BackgroundWithTargetRange from '../../components/trends/common/BackgroundWithTargetRange';
 import CBGSlicesAnimationContainer from './CBGSlicesAnimationContainer';
 import SMBGRangeAvgAnimationContainer from './SMBGRangeAvgAnimationContainer';
+import { NoData, dataTypes } from '../../components/trends/common/NoData';
```
jebeck: I'd really prefer if for consistency with the way other components are done if you stick to what I initially advised and just put a `displayTypes` in the default props. Makes use of all our components consistent (not having to import multiple pieces).

src/components/trends/common/NoData.js
```diff
@@ -0,0 +1,58 @@
...
+};
+
+NoData.defaultProps = {
+  displayTypes: { cbg: 'CGM', smbg: 'fingerstick' },
+  messageString: 'There is no <%= displayType %> data for this time period :(',
+};
+
+NoData.propTypes = {
+  messageString: PropTypes.string.isRequired,
+  dataType: React.PropTypes.oneOf(['cbg', 'smbg']),
```
jebeck: Can we just tack on `isRequired` here since that is how we expect to use it? It's great that it still displays a message even if we forget (as I saw in your tests), but in dev the `isRequired` will help by providing us with the React warning.

jebeck:
Can you address the final small comment here and merge this before I take another look at the render-days branch @jh-bate?

jh-bate:
done @jebeck - will now merge

PR #23, opened 2016-09-30
deps cleanup

jebeck:
My borking of the merge conflicts on package.json when I was trying to do some simple branch updating clued me in to some mis-categorized dependencies added for the settings work, so that is fixed here.

The tl;dr is: whenever you add a dependency that is used in actual viz code (not tests or tooling), ask yourself: "Is this also a dependency of blip?" (e.g., React)
- YES -> it goes in `devDependencies` **and** `peerDependencies` **and** gets added as an "external" in package.config.js
- NO -> it goes in `dependencies` **and** gets added as an "external" in package.config.js

I have also tried to explain this a bit in the README and would like it if everyone could take a look and LMK if it's clear and if we all understand: ping @krystophv @hntrdglss @jh-bate 

I could go into a lot more detail in the README, as I learned a lot more about how webpack externals and `npm install --production` works today, but I kept it pretty brief/concise for now. 😉

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,3 +1,4 @@
+coverage/
 data/
 src/
 stories/
--- a/.npmignore
+++ b/.npmignore
@@ -11,4 +11,5 @@ test/
 .travis.yml
 karma.conf.js
 loadtests.js
+package.config.js
 webpack.config.js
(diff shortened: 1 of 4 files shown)


jebeck:
[[viz] port smbg version of trends view to viz repo, Pt. 1](https://trello.com/c/35wkckjP/1-viz-port-smbg-version-of-trends-view-to-viz-repo-pt-1)

PR #26, opened 2016-10-06
Render days

jh-bate:


test/containers/trends/SMBGDaysContainer.test.js
```diff
@@ -0,0 +1,88 @@
...
+    unfocusSmbg: () => {},
+    xScale,
+    yScale,
+  };
+
+  before(() => {
+    wrapper = mount(<SMBGDaysContainer {...props} />);
+  });
+
+  describe('render', () => {
```
jebeck: Want to add a test re: not rendering anything if `!data`? Same goes for other components if such a test not already there...

src/redux/actions/trends.js
```diff
@@ -24,10 +24,10 @@ export function focusTrendsCbgSlice(userId, sliceData, slicePosition, focusedKey
   };
 }
 
-export function focusTrendsSmbg(userId, smbgData, smbgPosition) {
+export function focusTrendsSmbg(userId, smbgData, smbgPosition, smbgDay, smbgPositions, day) {
```
jh-bate: date `2016-10-20`
dayOfWeek `Thursday` 

works for me

jebeck:
Take note @jh-bate: when you added the keys where required in the latest commit, the problematic animation of the smbgs when you navigate across dates went away :D (Lesson: react-motion needs keys to keep track of which elements are which and probably default to array index if proper keys not provided...)

src/components/trends/common/Tooltip.js
```diff
@@ -0,0 +1,184 @@
...
+    const { tailWidth, tailHeight, borderWidth, borderColor, side } = this.props;
+    const tailSide = side === 'left' ? 'right' : 'left';
+    const padding = 10;
+    let marginOuterValue;
+    let marginInnerValue;
+    if (tailSide === 'left') {
+      marginOuterValue = `calc(-100% - 2 * ${padding}px)`;
+      marginInnerValue = `calc(-100% - (2 * ${padding}px - ${borderWidth + 1}px))`;
+    } else {
+      marginOuterValue = `calc(${padding}px + ${borderWidth}px)`;
```
krystophv: turns out pretty much everything we're targeting (I think) has `calc` support now: http://caniuse.com/calc/embed/

src/components/trends/common/Tooltip.js
```diff
@@ -0,0 +1,184 @@
...
+  }
+}
+
+Tooltip.propTypes = {
+  title: PropTypes.node,
+  content: PropTypes.node,
+  position: PropTypes.shape({
+    top: PropTypes.number.isRequired,
+    left: PropTypes.number.isRequired,
+  }),
```
krystophv: It's really not, technically. It'll just render at (0,0) relative to it's parent. I suppose it would probably make sense to require the prop, and make the containing component pass in the (0,0) explicitly if that's what they want.

src/components/trends/common/Tooltip.js
```diff
@@ -0,0 +1,184 @@
...
+      offset.top = topOffset + propOffset.top;
+      offset.left = leftOffset + propOffset.left;
+    }
+
+    this.setState({ offset });
+  }
+
+  renderTail(color = 'white') {
+    const { tailWidth, tailHeight, borderWidth, borderColor, side } = this.props;
+    const tailSide = side === 'left' ? 'right' : 'left';
```
krystophv: 👍

src/components/trends/common/Tooltip.js
jebeck: I think what you've just described is that it's not required for the component to not blow up, but it is pretty much required for how we expect the component to be used. So I think `isRequired` would be helpful, cuz then it gives users of the component React warning if they do it wrong.

src/redux/actions/trends.js
hntrdglss: 👍 I like that standardization.

jh-bate:
@jebeck yeah I totally expected that with the keys :)

jh-bate:
@krystophv @jebeck I think we are ready for a final round of reviews before we merge to the next branch :)

krystophv:
[[viz] port smbg version of trends view to viz repo](https://trello.com/c/35wkckjP/1-viz-port-smbg-version-of-trends-view-to-viz-repo)

src/components/trends/smbg/FocusedSMBGPointLabel.js
```diff
@@ -0,0 +1,134 @@
...
+      position={positions[i]}
+      side={'bottom'}
+      tail={false}
+      offset={{ top: SIMPLE_VALUE_TOP_OFFSET, left: 0 }}
+    />
+  ));
+  let focusedTooltip;
+  if (lines) {
+    focusedTooltip = (<Tooltip
+      title={<span className={styles.explainerText}>{smbgTime}</span>}
```
jebeck: As discussed, change `smbgTime` to something else - `lineDate` probably.

src/components/trends/common/Tooltip.css
```diff
@@ -0,0 +1,40 @@
+.tooltip {
+  display: block;
+  font-size: 14px;
```
jebeck: As mentioned in our meeting earlier, pull in updates from master and use the `defaultSize` from typography.css plz.

src/components/trends/common/Tooltip.css
```diff
@@ -0,0 +1,40 @@
...
+  background-color: #FFFFFF;
+  border: solid;
+  position: absolute;
+  border-radius: 4px;
+  opacity: 0.75;
+}
+
+.content {
+  display: flex;
```
jebeck: I can't remember if we had some flexbox styles separated out for composition in layout.css. Perhaps check?

src/components/trends/common/Tooltip.css
```diff
@@ -0,0 +1,40 @@
...
+  composes: mediumContrastText from '../../../styles/typography.css';
+  composes: content;
+  background-color: var(--tooltip-title-bg);
+}
+
+.tail {
+  width: 0px;
+  height: 0px;
+  border-style: solid;
+  border-color: transparent;
```
jebeck: OK, so as a check of my understanding...the `transparent` here is to cover all sides not overridden with a color in the JavaScript via higher-priority inline styles?

src/components/trends/common/Tooltip.js
```diff
@@ -0,0 +1,183 @@
...
+    if (tailSide === 'left') {
+      marginOuterValue = `calc(-100% - 2 * ${padding}px)`;
+      marginInnerValue = `calc(-100% - (2 * ${padding}px - ${borderWidth + 1}px))`;
+    } else {
+      marginOuterValue = `calc(${padding}px + ${borderWidth}px)`;
+      marginInnerValue = `${padding - 1}px`;
+    }
+    const borderSide = (tailSide === 'left') ? 'right' : 'left';
+
+    return (<div>
```
jebeck: Nitpick on style: I'd prefer if opening `<div>` was on next line.

I also think a simple comment about what each of the child `div`s does would be helpful - took me a while to figure out.

src/components/trends/smbg/FocusedSMBGPointLabel.js
```diff
@@ -0,0 +1,134 @@
...
+
+  const {
+    bgUnits,
+    focusedPoint: { data, position, dayPoints, positions },
+    timePrefs,
+    lines,
+  } = props;
+
+  let parsedTime;
+  if (timePrefs.timezoneAware) {
```
jebeck: This logic (or logic somewhat similar to it, after the WW data processing is in) is going to be needed a lot of places. Could you move it to the datetime utils and add tests?

src/components/trends/smbg/FocusedSMBGPointLabel.js
```diff
@@ -0,0 +1,134 @@
...
+  if (lines) {
+    focusedTooltip = (<Tooltip
+      title={<span className={styles.explainerText}>{smbgTime}</span>}
+      position={position}
+      side={'right'}
+      offset={{ top: SIMPLE_DAY_TOP_OFFSET, left: SIMPLE_DAY_LEFT_OFFSET }}
+    />);
+  } else {
+    focusedTooltip = (<Tooltip
+      title={<span className={styles.tipWrapper}>
```
jebeck: Just a note that as discussed we should be mindful of the potential need to abstract out a TooltipTitle and/or TooltipContent component(s) to avoid duplicating the `tipWrapper` styles in many places.

src/components/trends/smbg/FocusedSMBGPointLabel.js
```diff
@@ -0,0 +1,134 @@
...
+    positions: PropTypes.arrayOf(PropTypes.shape({
+      left: PropTypes.number.isRequired,
+      top: PropTypes.number.isRequired,
+    })).isRequired,
+  }),
+  grouped: React.PropTypes.bool.isRequired,
+  lines: React.PropTypes.bool.isRequired,
+  timePrefs: PropTypes.shape({
+    timezoneAware: React.PropTypes.bool.isRequired,
+    timezoneName: React.PropTypes.oneOfType([React.PropTypes.string, null]),
```
jebeck: Shouldn't there be an `isRequired` here too? Also is `null` really possible "in the wild"? I'm aware it's in a bunch of test fixtures, but those may be *unrealistic* test fixtures, which means the code is being modified to match test fixtures that won't occur in the wild 🙀

src/components/trends/smbg/FocusedSMBGRangeLabels.js
```diff
@@ -0,0 +1,101 @@
...
+
+const FocusedSMBGRangeLabels = (props) => {
+  const { focusedRange } = props;
+  if (!focusedRange) {
+    return null;
+  }
+
+  const { bgUnits, focusedRange: { data, position } } = props;
+  const timeFrom = millisecondsAsTimeOfDay(data.msX - ONE_AND_A_HALF_HOURS);
+  const timeTo = millisecondsAsTimeOfDay(data.msX + ONE_AND_A_HALF_HOURS);
```
jebeck: This breaks the flexibility to set the bin size to something other than three hours. Please follow the model [here](https://github.com/tidepool-org/viz/blob/master/src/utils/trends/data.js#L59) from the cbg code to add `msFrom` and `msTo` attributes based on the `binSize` in the `calculateSmbgStatsForBin` method.

src/components/trends/smbg/FocusedSMBGRangeLabels.js
```diff
@@ -0,0 +1,101 @@
...
+  const maxPosition = { top: position.yPositions.max, left: position.left };
+  const meanPosition = { top: position.yPositions.mean, left: position.left };
+  const meanSide = position.tooltipLeft ? 'left' : 'right';
+  const minPosition = { top: position.yPositions.min, left: position.left };
+  return (
+    <div className={styles.container}>
```
jebeck: Yes, this is the style I prefer for the JSX 😀

src/components/trends/smbg/FocusedSMBGRangeLabels.js
```diff
@@ -0,0 +1,101 @@
...
+        position={minPosition}
+        side={'bottom'}
+        tail={false}
+        offset={{ top: 5, left: 0 }}
+      />
+    </div>
+  );
+};
+
+FocusedSMBGRangeLabels.defaultProps = {};
```
jebeck: Remove since there's nothing in it? Maybe you intended to delete?

src/components/trends/smbg/SMBGDayLineAnimated.js
```diff
@@ -0,0 +1,107 @@
...
+import { line } from 'd3-shape';
+import _ from 'lodash';
+import cx from 'classnames';
+
+import { THREE_HRS } from '../../../utils/datetime';
+import { findBinForTimeOfDay } from '../../../utils/trends/data';
+
+import styles from './SMBGDayLineAnimated.css';
+
+const SMBGDayLineAnimated = (props) => {
```
jebeck: I'm getting this in the console, which I think we should figure out before this branch gets merged:
screenshot 2016-10-25 11 08 27

LMK if you need help with the detective work @jh-bate

src/containers/trends/SMBGDaysContainer.js
```diff
@@ -0,0 +1,135 @@
...
+ * == BSD2 LICENSE ==
+ */
+
+import React, { PropTypes } from 'react';
+import _ from 'lodash';
+
+import SMBGDayPointsAnimated from '../../components/trends/smbg/SMBGDayPointsAnimated';
+import SMBGDayLineAnimated from '../../components/trends/smbg/SMBGDayLineAnimated';
+
+const SMBGDaysContainer = (props) => {
```
jebeck: Since we've agreed to use "date" language for dates and "day" language for day-of-week (Mon, Tues, etc.), could we change the name of this to `SMBGsPerDateContainer` or something similar?

src/containers/trends/SMBGDaysContainer.js
```diff
@@ -0,0 +1,135 @@
...
+      value: PropTypes.number.isRequired,
+      localDate: PropTypes.string.isRequired,
+    })).isRequired,
+    smbgData: PropTypes.shape({
+      value: PropTypes.number.isRequired,
+    }),
+    smbgPosition: PropTypes.shape({
+      top: PropTypes.number.isRequired,
+      left: PropTypes.number.isRequired,
+    }),
```
jebeck: Should `smbgPosition` as a whole be required?

src/styles/colors.css
```diff
@@ -35,9 +35,18 @@
...
   --chrome: #989897;
 
   --text-medium-contrast: #727375;
   --text-high-contrast: #281946;
+
+  --bg-low: #FF8B7C;
+  --bg-target: #76D3A6;
+  --bg-high: #BB9AE7;
+
+  --smbg-dayline: var(--data-purple--medium);
```
jebeck: Why do you need another variable name for this? We just use `--data-purple--medium` directly in other places in the code (like the cbg side of trends view).

test/components/trends/smbg/FocusedSMBGPointLabel.test.js
```diff
@@ -0,0 +1,166 @@
...
+          focusedPoint={focusedPoint}
+          lines={false}
+          grouped
+          timePrefs={timePrefs}
+        />
+      );
+    });
+
+    it('should render individual point tooltips', () => {
+      expect(wrapper.find(formatClassesAsSelector(styles.number))).to.have.length(4);
```
jebeck: It strikes me as a bit fragile to count based on the styles, but I see that I set this precedent with the cbg slice tooltip tests, so, uh, ...`¯\_(ツ)_/¯` (Perhaps something we should all noodle on for the future, however.)

src/components/trends/smbg/SMBGDayLineAnimated.js
jh-bate: I have found the issue - remember the "strange" mapping function ... well funny story :) Just cleaning up the code and will commit it in the morning

jh-bate:
@krystophv I have addressed a number if the items with my commits above but still some work to do

src/components/trends/common/Tooltip.css
krystophv: There aren't yet, and there's no place that I could find that repeats more than a single flexbox style across our CSS yet, so there'd be no real benefit in making them composable classes (they'd have to be so granular right now that we'd just be repeating the CSS in a new way).

src/components/trends/common/Tooltip.css
krystophv: That's correct. `transparent` for all sides except the one with the color nets the triangle of color that forms the 'tail'. The element's center is actually at the most acute point of the triangle.

src/components/trends/smbg/FocusedSMBGPointLabel.js
krystophv: In the case that we are in timezone naive mode, the `timezoneName` is not provided and React will throw a warning. Actually, this is what's currently happening in the TrendsContainer in timezone naive mode because it has that property marked as `isRequired`. (https://github.com/tidepool-org/viz/blob/render-days/src/containers/trends/TrendsContainer.js#L65). So we might need to consider making it not `isRequired` everywhere or change the state handling so we always provide something for `timezoneName`, even in timezone-unaware mode. (Or maybe a `.oneOf([PropTypes.shape({}), PropTypes.shape({})])` which specifies both potential shapes, but that seems a little convoluted).

jebeck: Ah, OK, I stand corrected. Let's use this version everywhere then I guess, like if you could stick it in the TrendsContainer and anywhere else that does it wrong, that would be great, thanks!

krystophv:
@jebeck - design review revisions (https://github.com/tidepool-org/viz/compare/fb5eeec...673c7a0) are in.

jebeck:
I'm getting React warnings sometimes related to tooltips, but having trouble pinning down the exact circumstances, though it appears to be more common when I toggle from CGM to BGM trends. Could you try to look into the first one a bit @krystophv?
screenshot 2016-11-08 16 18 54

The second one is due to `tooltipLeftThreshold` missing in the props passed to `SMBGDayLineAnimated` [here](https://github.com/tidepool-org/viz/blob/render-days/src/containers/trends/SMBGsByDateContainer.js#L48) (compare with the instance below it), so either @krystophv or @jh-bate could fix that.

I just got this one too, so I think another look at the props for `FocusedSMBGPointLabel` makes sense:
screenshot 2016-11-08 16 33 04

src/components/trends/smbg/FocusedSMBGRangeLabels.js
jebeck: Perhaps you didn't see this comment before? Can you remove the unused `defaultProps` here?

src/components/trends/smbg/FocusedSMBGRangeLabels.js
```diff
@@ -0,0 +1,101 @@
...
+      yPositions: PropTypes.shape({
+        max: PropTypes.number.isRequired,
+        mean: PropTypes.number.isRequired,
+        min: PropTypes.number.isRequired,
+      }).isRequired,
+    }).isRequired,
+  }),
+  timePrefs: PropTypes.shape({
+    timezoneAware: React.PropTypes.bool.isRequired,
+    timezoneName: React.PropTypes.oneOfType([React.PropTypes.string, null]),
```
jebeck: Have you done a pass through on all the code (outside of just smbg trends) as requested above? Would love it if we're confident we're improving things as we go.

jebeck:
@krystophv: One of the things I pointed out before where feature parity with existing tideline smbg Trends view had not been achieved does not appear to have been addressed. Has it been discussed with @skrugman? It's this one: https://tidepoolteam.slack.com/files/jebeck/F2U2V0HFT/screenshot_2016-10-25_11.05.04.png

If it hasn't been discussed, please do that!

src/components/common/tooltips/Tooltip.js
```diff
@@ -0,0 +1,190 @@
...
+      const tailCenter = {
+        top: tailRect.top + (tailRect.height / 2),
+        left: tailRect.left + (tailRect.width / 2),
+      };
+      offset.top = -tailCenter.top + tooltipRect.top + propOffset.top;
+      offset.left = -tailCenter.left + tooltipRect.left + propOffset.left;
+    } else {
+      let leftOffset;
+      let topOffset;
+      switch (this.props.side) {
```
jebeck: Could you add some tests around this logic? I think it'd be useful to prevent future regressions to know what the offsets should be (and also as pseudo-documentation/examples of how the offsets work).

src/components/trends/smbg/FocusedSMBGRangeLabels.js
krystophv: Yeah, everywhere that `timezoneName` is in propTypes.

PR #27, opened 2016-10-22
device settings revisions

jebeck:
The iterations, specifically, are as follows:
- alignment of column titles and text on non-Tandem
- title bars taller & with background (Tandem & non-Tandem)
- more space between titles & units
- fix for basal description overflow problem: "Active on upload" on a different line
- layout changes for non-Tandem (more space between basal & bolus groups; better wrapping of bolus sub-sections)
- bigger font on non-Tandem basal & bolus section titles
- capitalize Medtronic schedule names
- Tandem: unbold leftmost column time labels
- audit & adjust (if needed) all device-specific language
- restore ability to select & copy text (reported support issue) (via dependent [blip PR #341](https://github.com/tidepool-org/blip/pull/341)) 
- allow `medtronic` key in addition to `carelink` so that direct read from MedT pumps WIP can see pump settings display

In terms of code change, the biggest thing I had to do to accomplish these things was consolidate all the non-Tandem components into a single component, moving the specifics (just string constants) to the "factory". (This was necessary because some of the changes required structural change, and that was very repetitive to try to accomplish across 3 nearly identical components.)

Ping @jh-bate for review, or @krystophv if @jh-bate doesn't have time for the review on NZ Monday.


jebeck:
[[blip] [device settings] UI improvements for new device settings [dev:2 tst: 1, ux:1 ]](https://trello.com/c/3orasmyw/40-blip-device-settings-ui-improvements-for-new-device-settings-dev-2-tst-1-ux-1)

src/components/settings/common/Table.js
```diff
@@ -64,12 +79,13 @@ class Table extends React.Component {
     let tableContents = [];
 
     if (this.props.title) {
+      const { label: { main, secondary } } = this.props.title;
       const title = (
         <caption
-          key={this.props.title.label}
+          key={main}
           className={this.props.title.className}
```
jh-bate: seems we should then also be pulling out the `className` from the props title

src/components/settings/nontandem/NonTandem.js
```diff
@@ -0,0 +1,250 @@
...
+ */
+
+import React, { PropTypes } from 'react';
+import _ from 'lodash';
+
+import Header from '../common/Header';
+import Table from '../common/Table';
+import CollapsibleContainer from '../common/CollapsibleContainer';
+
+import * as constants from '../../../utils/constants';
```
jh-bate: should it actually just be the specific constants?

src/utils/settings/data.js
```diff
@@ -130,20 +129,15 @@ export function getTotalBasalRates(scheduleData) {
  * getScheduleLabel
  * @param  {String} scheduleName basal schedule name
  * @param  {String} activeName   basal name active at upload timestamp
- * @param  {String} activeClass  class name to apply to active text
- * @param  {String} units        units to append to schedule name
  *
- * @return {String}              formatted basal schedule label
+ * @return {String}              object representing basal schedule label
```
jh-bate: should this then be `@return {Object}`

stories/components/settings/medtronic/Medtronic.js
```diff
@@ -2,7 +2,8 @@ import React from 'react';
 
 import { storiesOf } from '@kadira/storybook';
 
-import Medtronic from '../../../../src/components/settings/medtronic/Medtronic';
+import { getChart } from '../../../../src/utils/settings/factory';
+const Component = getChart('carelink');
```
jh-bate: why not just call it what it is rather than a generic `Component`?

stories/components/settings/omnipod/Omnipod.js
```diff
@@ -2,18 +2,19 @@ import React from 'react';
 
 import { storiesOf } from '@kadira/storybook';
 
-import Omnipod from '../../../../src/components/settings/omnipod/Omnipod';
+import { getChart } from '../../../../src/utils/settings/factory';
+const Component = getChart('insulet');
```
jh-bate: again why the generic name?

src/components/settings/common/Table.js
jebeck: I don't understand your comment here. Can you elaborate?

jh-bate: so when looking at the whole file just a few lines above 

```
const { label: { main, secondary } } = this.props.title;
```

as we are already pulling stuff from the title prop would it not make sense to also get the `className` at the same time ?

jebeck: Ah, OK, sure.

src/components/settings/nontandem/NonTandem.js
jebeck: I changed it. I think I may have c&ped from another file where it was done the other way, so at some point we may want to do a pass for consistency...

src/utils/settings/data.js
jebeck: D'oh, good catch. I did update the description! Just not the type /o\

stories/components/settings/medtronic/Medtronic.js
jebeck: I didn't like calling it a "chart" because it's not a chart. But I can change it to `Settings` or `SettingsComponent`?

jebeck:
I believe I've addressed all feedback @jh-bate. Mergeable?

PR #31, opened 2016-10-27
persist state of open/collapsed basal schedules (or settings profiles, for Tandem devices)

jebeck:
This connects the device settings up to the redux state so that we can persist the basal schedule (or Tandem settings profile) that the user has opened or collapsed. We still default to the active schedule as opened, however.

Perhaps @hntrdglss is best to review at the moment? I wouldn't recommend stepping through the commits in this case, as I changed my mind about the interface (particularly wrt to the PumpSettingsContainer) about halfway through... Rather I'd recommend just looking at the changes as a whole.

There will be a companion blip PR to this, but I've still got a test to fix.


jebeck:
[persist state on expanded/collapsed basal schedules and Tandem timed settings profiles when navigating across views (daily, trends, &c) [viz] [device settings] [dev: 2 ux: 1 tst:1]](https://trello.com/c/a7kVgvGO/43-persist-state-on-expanded-collapsed-basal-schedules-and-tandem-timed-settings-profiles-when-navigating-across-views-daily-trends)

src/containers/settings/PumpSettingsContainer.js
```diff
@@ -0,0 +1,115 @@
...
+          manufacturerKey={manufacturerKey}
+          openedSections={settingsState[manufacturerKey]}
+          pumpSettings={pumpSettings}
+          timePrefs={timePrefs}
+          toggleBasalScheduleExpansion={toggleFn}
+        />
+      );
+    }
+    // eslint-disable-next-line no-console
+    console.warn(`Unknown manufacturer key: [${manufacturerKey}]!`);
```
hntrdglss: Just wanted to make sure you'd like to keep this in production.

jebeck: Yes, I thought it'd be a good thing to keep.

PR #32, opened 2016-11-07
fix MedT active at upload regression

jebeck:
Discovered a regression (missing 'Active at upload' text for MedT) after briefly deploying blip v0.14.11 to prd before the offsite. (Rolled back to v0.14.8 after discovering it.)

The root cause of the dropped 'Active at upload' text was basically some incorrect fixture data for `pumpSettings` with capitalized basal schedule names (which **never** happens with MedT pumps in the wild because basal schedule names are not under user control). This masked how adding capitalization to MedT broke the logic to detect matching between each schedule name at the active schedule. So this PR fixes the underlying data fixtures and (after consultation with BA) only capitalizes MedT basal schedule names and preserves user-controlled capitalization on other pumps (tests have been added to prove this).

@hntrdglss could you review ASAP so we can clear the way for other changes that are almost ready for testing?

### DIFF
--- a/data/pumpSettings/animas/flatrate.json
+++ b/data/pumpSettings/animas/flatrate.json
@@ -1,9 +1,9 @@
 {
   "type": "pumpSettings",
-  "activeSchedule": "Normal",
+  "activeSchedule": "normal",
   "basalSchedules": [
     {
-      "name": "Normal",
+      "name": "normal",
       "value": [
         {
           "start": 0,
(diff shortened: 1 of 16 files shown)


jebeck:
[[blip] [device settings] UI improvements for new device settings [dev:2 tst: 1, ux:1 ]](https://trello.com/c/3orasmyw/40-blip-device-settings-ui-improvements-for-new-device-settings-dev-2-tst-1-ux-1)

PR #34, opened 2016-11-11
Port smbg trends

krystophv:
The net changes for porting SMBG trends views to the viz repo


krystophv:
[[viz] port smbg version of trends view to viz repo](https://trello.com/c/35wkckjP/1-viz-port-smbg-version-of-trends-view-to-viz-repo)

krystophv:
[[viz] port smbg version of trends view to viz repo](https://trello.com/c/35wkckjP/1-viz-port-smbg-version-of-trends-view-to-viz-repo)

krystophv:
ping @jebeck - I believe this is ready for review

jebeck:
Crashes when I navigate to device settings locally @krystophv - I suspect you haven't merged `port-smbg-trends` from master in blip to update.

krystophv:
@jebeck - that was correct. the corresponding blip branch has been updated from master.

src/containers/trends/TrendsContainer.js
```diff
@@ -199,6 +199,9 @@ export class TrendsContainer extends React.Component {
       mostRecent: mostRecent.toISOString(),
       xScale: scaleLinear().domain([0, 864e5]),
       yScale,
+      selectDay: (date) => {
+        this.props.onSelectDay(datetime.midDayForDate(date, timePrefs));
+      },
```
jebeck: Why does this need to be in state? Can't we just define it as an instance method?

test/utils/datetime.test.js
```diff
@@ -332,4 +332,24 @@ describe('datetime', () => {
         .to.throw('time and deviceTime must be a ISO-formatted String timestamp');
     });
   });
+
+  describe('midDayForDate', () => {
```
jebeck: Please add more test cases here to prove that this behaves as we desire for the edge cases of the day that we switch to DST and from DST. The goal is that in the case of the foreshortened day (23 hrs), you still see midnight on the left edge (-432e5 milliseconds from result, not 12 "clock" hours). Ditto for the lengthened day (25 hrs).

Also as a tip: it's generally best to use a timezone in tests that none of us is in. This prevents tests that are passing only due to the fact that you're in the timezone you're using in the tests. They then end up failing on Travis, causing much 😭. Obviously that didn't happen here, but just as a tip. [I often use Mountain or Central time (or CEST) to guard against this...]

src/containers/trends/TrendsContainer.js
krystophv: It can be, we just have to execute a function that returns a function that we can pass down so the `this` reference gets bound at this class instance.

PR #36, opened 2016-12-01
CGM Trends Range Selections

krystophv:
Allows the user to toggle the display of individual ranges in the CGM Trends view.


krystophv:
[CGM Trends 1 - Add Checkboxes, default top/bottom off [dev: 5, tst: 1, ux: 1]](https://trello.com/c/SoqZj1GH/52-cgm-trends-1-add-checkboxes-default-top-bottom-off-dev-5-tst-1-ux-1)

krystophv:
Is dependency of: https://github.com/tidepool-org/blip/pull/363

krystophv:
ping @jebeck for review (or feel free to pass to next available person)

src/components/trends/cbg/RangeSelect.js
```diff
@@ -0,0 +1,87 @@
...
+    turnOnCbgRange: _.partial(
+      actions.turnOnCbgRange, ownProps.currentPatientInViewId
+    ),
+    turnOffCbgRange: _.partial(
+      actions.turnOffCbgRange, ownProps.currentPatientInViewId
+    ),
+  }, dispatch);
+}
+
+export default connect(
```
jebeck: A couple of comments here: the first and third functions can just be absent (can pass `null` for the first and just leave the third out - if you were looking at the `TrendsContainer` example, it adds an additional property in the third function (`{withRef: true}`) for compatibility with blip's parent-child component communication through refs (an outdated strategy), but nothing like that is necessary here). Generally `connect()`-ed components are also considered containers, not components.

Back in the baby days of redux, it wasn't considered very good practice to nest a lot of `connect()`-ed components, but apparently [the advice has changed a little](https://github.com/reactjs/react-redux/issues/159#issuecomment-148510079) (linking for Dan's comment, his link is now broken). However, it does sound like there are still [some potential performance issues](https://github.com/reactjs/redux/issues/125), for which there is [a potential middleware solution](https://github.com/acdlite/redux-batched-updates).

Since we're starting to be concerned about performance here in viz, I think it'd be good to discuss pros & cons of the options here and sync up on future strategy (with @hntrdglss as well). Also more generally the approach you've taken here is not what I expected. Since we already had code in blip that renders checkboxes for the BGM options and since the checkboxes have to be rendered in blip code due to the current split between what blip controls of the dataviz options, I expected that code to be reused (and there was a comment about this on the card ("...the checkboxes should use the same code as the checkboxes for the options available on BGM trends (Lines on/off, Range & Avg on/off, etc.)" from Nov. 8 when @darinkrauss was helping to get things more precise on the card). I get that the Trello cards are full of lots of info in a not easily digestible format, so I'm not surprised you missed that, but it might have been nice to talk through the approach before the PR was up to make sure we're all on the same page about approach and have any discussions about pros & cons of options with respect to future code.

krystophv:
@jebeck - addressed the CSS issue and leaving the `connect()`-edness (outside of fixing the args) till we have a larger discussion. I think this is back up for review.

PR #37, opened 2016-12-21
Tooltip component updates

jebeck:
The changes here add four main features to the Tooltip component:
- the Tooltips are now invisible to pointer events, to avoid "stuttering" if connection between the original hover target and the pointer is broken due to hovering on a bit of the tooltip itself!
- the total offset gets recalculated from new props if the component is updated (via the `componentWillReceiveProps` lifecycle hook)
- ability to pass in a `backgroundColor` in props
- ability to pass in a `horizontal` instead of `left` in the `offset` prop: if/when you do this, the calculated total offset will flip the sign of the value specified in `horizontal` when the tooltip is on the left...this avoids repeating [logic such as found here](https://github.com/tidepool-org/viz/blob/master/src/components/trends/smbg/FocusedSMBGPointLabel.js#L76) all over the damn place

The `horizontal` option is an *add* and the behavior of passing in `left` remains unchanged (i.e., backwards-compatible) but I did update the components that currently passed in a `left` where they would benefit from `horizontal` because of the sign-flipping logic.

Added tests for all of the above, of course. Had a bit of trouble with the lifecycle hooks tests and will comment inline about what I did.

### DIFF
--- a/test/components/common/tooltip/Tooltip.test.js
+++ b/test/components/common/tooltip/Tooltip.test.js
@@ -73,7 +73,7 @@ describe('Tooltip', () => {
     const tooltipElem = wrapper.node.element;
     const tooltipgetBoundingClientRect = sinon.stub(tooltipElem, 'getBoundingClientRect');
     tooltipgetBoundingClientRect.returns({ top: 50, left: 50, height: 100, width: 100 });
-    wrapper.instance().calculateOffset();
+    wrapper.instance().calculateOffset(wrapper.props());
     expect(wrapper.state()).to.deep.equal({ offset: { top: -5, left: -5 } });
   });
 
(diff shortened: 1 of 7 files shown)


jebeck:
[<ux: 2 dev:8 tst:2> CGM Trends 2b - Implement exposing cbgs inside slices & hover to see full CGM date trace](https://trello.com/c/AefCI0rV/82-ux-2-dev-8-tst-2-cgm-trends-2b-implement-exposing-cbgs-inside-slices-hover-to-see-full-cgm-date-trace)

test/components/common/tooltip/Tooltip.test.js
```diff
@@ -160,7 +160,125 @@ describe('Tooltip', () => {
...
+        <Tooltip
+          position={position}
+          title={title}
+          content={content}
+        />
+      );
+      expect(Tooltip.prototype.componentDidMount.calledOnce).to.be.true;
+    });
+
+    it('componentDidMount in turn calls calculateOffset', () => {
```
krystophv: Is there something wrong with treating `calculateOffset` the same way that `componentDidMount` is in the test above this one? (like `sinon.spy(Tooltip.prototype, 'calculateOffset');`) It seems to work as expected for me (unless I'm missing something obvious, which is completely possible) as long as I `.reset()` the spy for each successive test.

jebeck: Fixed. Just a silly brain fart - was fixated on pulling `calculateOffset` off the mounted instance rather than the prototype...

PR #38, opened 2017-01-06
cbg slice changes for CGM trends new feature work

jebeck:
This is the first of quite of few PRs in the larger project (on the feature branch jebeck/trends-expose-cgm-date-traces) to expose CGM traces as part of the hover interaction on each segment of a CGM trends "slice." 

There are a whole bunch of other user experience changes that are coming along with this project, including many styling changes. [This Zeplin link](https://app.zeplin.io/project.html#pid=57b23ddc945171152c99fab2&sid=586d6821d7a1826f18712049) is a pretty good reference for the styling changes to expect.

This pull request contains the changes to the cbg "slice" rendering code, including:
- reorganizing the use of TransitionMotion for performance (mainly) as well as adding more animations (on initial mount, slice exit, and slice re-entry)
- splitting the three layered components of the "slice" for each time bin (100%, 80%, 50% of data) into five instead (top 10, bottom 10, upper 15, lower 15, inner 50%), without layering (because this will be required for new interactions)

Since the median is now colored according to relationship with the target BG range, there's also a new utility (& tests) for determining a particular BG value's relationship to the target range.

Note that hover labels are gone for the moment - the updates to that part will be the next PR. (It felt like what's here was plenty for a review.)

Looking for feedback in particular on whether I should be `connect()`ing in different places...


jebeck:

test/components/trends/cbg/CBGSliceAnimated.test.js
```diff
@@ -0,0 +1,263 @@
...
+      expect(wrapper.find(`#cbgSlice-${datum.id}`).length).to.equal(1);
+      expect(wrapper.find('rect').length).to.equal(6);
+    });
+
+    it('should render the #median rect on top (i.e., last)', () => {
+      expect(wrapper.find('rect').last().prop('id')).to.equal('median');
+    });
+
+    it('should vertically center the median rect on the value', () => {
+      const medianRect = wrapper.find('rect').last();
```
krystophv: Travis is unhappy about this const never being used.

jebeck: Whoops, I can fix that in a min. Thanks!

test/utils/constants.test.js
```diff
@@ -1,5 +1,19 @@
-/* eslint-env node, mocha */
-/* eslint no-console: 0*/
+/*
+ * == BSD2 LICENSE ==
+ * Copyright (c) 2016, Tidepool Project
```
krystophv: Adding the copyright might be better in a dedicated repo-encompassing PR where it's added/modified to all related files instead of a code-change PR, but if it's gonna be here, we might as well make it 2017, right?

src/components/trends/cbg/CBGSliceAnimated.js
```diff
@@ -0,0 +1,244 @@
...
+    super(props);
+    this.willEnter = this.willEnter.bind(this);
+    this.willLeave = this.willLeave.bind(this);
+  }
+
+  getDefaultYPosition(bgBounds, yScale) {
+    const { targetLowerBound, targetUpperBound } = bgBounds;
+    // default Y position is the center of the target range
+    // i.e., 100 mg/dL if target range is 80-120 mg/dL
+    return yScale(targetUpperBound - (targetUpperBound - targetLowerBound) / 2);
```
krystophv: Is this something that could be calculated once per `props` update and stored in `state`? Or, if having this component be stateless is really preferred, maybe a `_.memoize`ation?

src/containers/trends/TrendsSVGContainer.js
```diff
@@ -105,20 +109,13 @@ export class TrendsSVGContainer extends React.Component {
 
   renderCbg() {
     if (this.props.showingCbg) {
```
jebeck: Nope! One of the <em>really</em> important things about using TransitionMotion to do the mount, enter, and exit animations, is that the TransitionMotion container has to still be rendered in order for those "transition" animations to happen. So the smbg side still has the check because I haven't updated the TransitionMotion usage there (yet). Now the check has moved lower - <em>inside</em> TransitionMotion [here](https://github.com/tidepool-org/viz/blob/jebeck/trends-cbgslice-changes/src/components/trends/cbg/CBGSliceAnimated.js#L216-L218).

src/utils/bloodglucose.js
```diff
@@ -0,0 +1,42 @@
...
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+import _ from 'lodash';
+
+/**
+ * classifyBgValue
+ * @param {Object} bgBounds - object describing boundaries for blood glucose categories
+ * @param {Number} bgValue - integer or float blood glucose value in either mg/dL or mmol/L
```
jebeck: Ah, sure, can do.

test/utils/constants.test.js
jebeck: Well this file was created in 2016, but it just happened to be missing the header, and I noticed. ¯\_(ツ)_/¯

src/components/trends/cbg/CBGSliceAnimated.js
jebeck: The advice from the react-motion docs is to use instance methods for the `willLeave` and `willEnter` functions, and that makes sense to me given what I've read about React & perf stuff (potential gotcha with things like arrow functions in props because it's a new (update-forcing) function every render cycle). So assuming that stateless components are *not* an option for TransitionMotion-animated components of this kind, does that change your feedback on this `getDefaultYPosition` method? It didn't strike me as particularly problematic, but maybe I'm missing something.

If we derived the value from props and stored it in state we'd need to implement `componentWillReceiveProps` and that seems like premature optimization before we even have the ability to update target range or scales, but OTOH it also seems like a bad React practice to leave it without a `componentWillReceiveProps` (a bug waiting to happen).

src/utils/bloodglucose.js
jebeck: And done.

src/components/trends/cbg/CBGSliceAnimated.js
jebeck: Actually now that I'm working on the smbg side I'm realizing I'll need this in more places, so I'm thinking a simple higher-order component is the way to go (thinking about the advice in the recent [Mixins Considered Harmful](https://facebook.github.io/react/blog/2016/07/13/mixins-considered-harmful.html)). Thoughts? It'll be in the next PR if you agree.

src/containers/trends/TrendsSVGContainer.js
krystophv: 👍

src/components/trends/cbg/CBGSliceAnimated.js
krystophv: Yeah, a higher-order component seems like a pretty good solution.

PR #40, opened 2017-01-11
updates on the smbg side of Trends view

jebeck:
This branch contains:

1. Styling & rendering updates on the smbg side of Trends view to align with new slices on the cbg side (i.e., grays and coloring mean by BG category + making it rectangular)
1. It also refactors the rendering components in line with the work on the CGM side for more transitional animations
1. To achieve (2) some interesting stuff had to be added, and I hope to work on preparing some documentation/guides/simple examples soon (perhaps on my flights this weekend). The core idea is that in order to achieve *exit* animations, you need to still be rendering a `TransitionMotion` component with the same key as the thing exiting. This means that in order to exit individual smbgs when you page through dates, you need to still render a `TransitionMotion` for each exited date. So you'll see changes in `TrendsContainer` (and a bit in blip) to track the `prevDateDomain` in state and then a new utility to calculate all dates from a union of the previous and current date ranges so that we can render a `TransitionMotion` container for every current *and* exiting thing.
1. As discussed with @krystophv in #38 there's also a HOC here for calculating and passing through a `defaultY` parameter (for use in `defaultStyles` and `willEnter` and `willLeave` hooks for `TransitionMotion`). A question here: I think I've seen people using camelCase for HOC names instead of CapitalizedCase... Do we like this? (<strike>Oops, just noticed I still need to add tests for the HOC, will do that v. soon!</strike> Done!)
1. Renamed all the inconsistent use of 'day' to 'date'. Reminder: we are reserving 'day' for only referring to days of the week (Monday, Tuesday, etc.). On the smbg side of Trends we render stacked *dates* of individual smbgs optionally connected by a line within each *date*.

(I semi-apologize for all the places where I alphabetized props... I can't help it. Though I believe it really does help me keep track of things when there are a lot. Can I convince @krystophv and @hntrdglss that this should be part of our style??)

See also [blip #375](https://github.com/tidepool-org/blip/pull/375).


jebeck:

src/containers/trends/TrendsSVGContainer.js
```diff
@@ -48,9 +48,8 @@ import { THREE_HRS } from '../../utils/datetime';
 import Background from '../../components/trends/common/Background';
 import CBGSlicesContainer from './CBGSlicesContainer';
 import SMBGsByDateContainer from './SMBGsByDateContainer';
-import SMBGRangeAvgAnimationContainer from './SMBGRangeAvgAnimationContainer';
-import SMBGAvg from '../../components/trends/smbg/SMBGAvg';
```
krystophv: Here we drop the old SMBGAvg component, but it looks like we don't pick up a new one and (as far as I can tell) we're not displaying one.

jebeck: Yeah this is still an open question after our meeting last Thursday, and we're going to revisit once the rest of the new feature work is ready for design review. With the styling changes the average was really distracting, and the info appears on hover, so as a possible temporary or interim solution we've decided to just remove it. I thought I had updated the PR description to make a note of this, but I see I didn't. Sorry! Must have forgot.

I've actually made a notes of the commits to revert if/when we bring the average back: 316f24b985d0401554dab0d30e741ec1346810a5 here in viz and 72395fa4eaada1e0c073d435e5ea9370055568cc in blip.

jebeck:
Thanks @krystophv! I apologize about the huge mix of changes. I definitely need to work on that and will try in the future.

PR #43, opened 2017-01-17
trends update cbg slice hover

jebeck:
This updates the cbg slice (or rather, segment of a slice) hover to use a new Tooltip-based module which is also refactored from the smbg version to be common to both.

See also [blip #376](https://github.com/tidepool-org/blip/pull/376).


jebeck:

jebeck:
Intentional. Currently I set the median to `pointer-events: none;` so that it's invisible and doesn't interfere with the interquartile segment hover. So there is a bit of redundant code stopping the median from producing any hover text. I added those tests because something like that existed before, and for a sort of documentation, but if they add more confusion than help LMK and I can delete them.

PR #44, opened 2017-01-19
show cbg date traces on segment hover

jebeck:
This adds a bunch of new components (and tests for them) so that when you hover on one of the five segments of a cbg slice in the CGM version of Trends view, all the sensor traces for the dates that intersect with that segment appear on top of the slices display.

I didn't put any tests around the special animation lifecycle methods (added by `TransitionGroupPlus`) where we're doing the GSAP-based animation because it seemed like any testing there would basically be either testing `TransitionGroupPlus` or GSAP itself, not our code... LMK if you feel differently.

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,10 +99,12 @@
     "d3-scale": "1.0.3",
     "d3-shape": "1.0.3",
     "d3-time": "1.0.2",
+    "gsap": "1.19.0",
     "react-collapse": "2.3.1",
     "react-dimensions": "2.0.0-alpha1",
     "react-height": "2.1.1",
-    "react-motion": "0.4.4"
+    "react-motion": "0.4.4",
+    "react-transition-group-plus": "0.4.0"
   },
   "peerDependencies": {
     "classnames": "2.x",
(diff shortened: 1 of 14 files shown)


jebeck:

jebeck:
I believe there's an issue with `TransitionGroupPlus` that's causing errors in the console: see [here](https://github.com/cheapsteak/react-transition-group-plus/issues/12) for details.

jebeck:
Huh, I see what you mean about the interaction after you unselect things, but I'm confused about why we're still rendering them after they exit. I thought `TransitionMotion` would keep them around long enough to exit them but then actually unmount them... [That's what the docs say.](https://github.com/chenglou/react-motion#transitionmotion-)

In any case, setting the pointer events to `none` in the `willLeave` should do it. I also have a hunch about maybe what I'm doing wrong in the `willLeave`...

jebeck:
Going to merge now to continue with feature work for design review meeting tomorrow, but will address the issue of the non-disappearing rectangles in a future PR.

PR #45, opened 2017-01-24
add ability to "stick" date traces on cbg slice segment click

jebeck:
This is the next feature add in the Trends work - allowing the user to "stick" the CGM sensor traces for the dates that intersect with a given slice segment so that they stay even when the user stops hovering on that segment. This way the user can build up a set of traces for comparison.

Clicking anywhere on the chart background "unsticks" all traces.

There will be another tiny PR on this with some alphabetization - I find it really useful to alphabetize the redux action creators and reducers, mainly so that going back and forth between the source and test files I can easily tell which things do & don't have tests.

Testing the click interaction on the SVG for unsticking was tough because we really need to use `mount` to test each code path fully but we can't because of the issues with `TransitionMotion` and full mount... Ideas welcome!

### DIFF
--- a/test/containers/trends/TrendsSVGContainer.test.js
+++ b/test/containers/trends/TrendsSVGContainer.test.js
@@ -61,7 +61,11 @@ describe('TrendsSVGContainer', () => {
     // normally provided by react-dimensions wrapper but we test w/o that
     containerWidth: 960,
     dates: [],
-    cbgData: [{ id: 'a2b3c4', localDate: '2017-01-01', msPer24: 6000, value: 180 }],
+    cbgData: [
+      { id: 'foo', localDate: '2017-01-01', msPer24: 6000, value: 180 },
+      { id: 'bar', localDate: '2017-01-05', msPer24: 5000, value: 60 },
+      { id: 'baz', localDate: '2017-01-05', msPer24: 1000, value: 175 },
+    ],
     smbgData: [{ id: 'a2b3c4', localDate: '2016-07-04', msPer24: 6000, value: 180 }],
     displayFlags: {
       cbg100Enabled: false,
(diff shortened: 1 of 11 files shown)


jebeck:

test/containers/trends/TrendsSVGContainer.test.js
```diff
@@ -330,5 +391,24 @@ describe('TrendsSVGContainer', () => {
...
+          // we have to stub here instead of spying because we're not doing a full mount
+          // and so the click handler's inspection of the event target will fail w/a spy
+          const clickStub = sinon.stub(TrendsSVGContainer.prototype, 'handleClick');
+          const stuckDatesProps = _.assign(
+            {}, props, { stuckCbgDateTraces: ['2017-01-01'] },
+          );
+          const stuckDatesWrapper = shallow(<TrendsSVGContainer {...stuckDatesProps} />);
+          expect(clickStub.callCount).to.equal(0);
+          const svg = stuckDatesWrapper.find('svg');
+          svg.simulate('click');
```
krystophv: We can still use a `sinon.spy` here if we pass in an object for the click event. like:
`svg.simulate('click', { target: { id: 'cbgSlice' } });`
see: https://github.com/airbnb/enzyme/blob/master/docs/api/ShallowWrapper/simulate.md for details.

jebeck: Ahhh, very good to know! If we were keeping this I'd update, but it looks like we won't be...

jebeck:
We've decided to wait on this interaction, so closing. I won't delete the branch right away in case any of this is useful if we come back to it relatively soon.

test/containers/trends/TrendsSVGContainer.test.js
jebeck: Referred back to this for new tests I'm writing. Thanks again @krystophv

PR #49, opened 2017-01-27
add interactions to CGM date traces

jebeck:
Namely:
- hover on a date trace surfaces a date label (via [blip #381](https://github.com/tidepool-org/blip/pull/381))
- click on a date trace goes to Daily view for that date

One of the additional changes here was that I changed the default date format for `formatDisplayDate` from the format used in settings to the format now desired to be used across BGM and CGM Trends for long-format dates. The fact that a bunch of settings tests failed when I did this made me very happy, as it was then easy to avoid a regression in the format there 😎


jebeck:

PR #50, opened 2017-02-01
performance improvements for Trends

jebeck:
To summarize the changes here:

- As with [blip #382](https://github.com/tidepool-org/blip/pull/382), this PR includes an update to React 15.4.x so that we can use the new `PureComponent` (instead of `Component`) wherever possible (which implements a default `shouldComponentUpdate` with a shallow comparison). [As I made the changes here, I used the React perf tools quite a bit and dropping `PureComponent` in was the first big win.]
- I factored out the median component from the rest of the cbg slice because it's no longer interactive and there was a lot of if/else logic that seemed unnecessarily complex when it was part of the CBGSliceAnimated.
- All the low-level rendering components for interactive SVG elements (with hover and/or click behaviors) have been refactored to be `connect()`ed so that they can pull their action creator functions in directly (avoiding rerenders from the `_.partial` we were using every render cycle to bind the `userId` in as well as occasional other places we were creating new arrow functions). I followed [the advice from Benchling here](http://benchling.engineering/performance-engineering-with-react/#functionbindinlineanonymousfunctions) on this issue, although in a few cases I still had to leave an arrow function in; this seemed perfectly acceptable where I did it (only on actual rendering of DOM elements) because there were no sub-trees renders to worry about.
- I fixed all the places where I was implementing `willEnter` and more problematically `willLeave` incorrectly for `TransitionMotion` resulting in the components not exiting after animating out (as you previously pointed out to me @krystophv - great catch!). The problem was that I was defining too many styles in the object(s) returned in those methods: it's important to return *only* the properties you are animating in those methods (because something is checking that the final values get reached and only *then* unmounting the component), and the `render` will still have access to any other non-animated properties on the `style` object. Not that the docs explain this or anything 😐

Related to the third point: an update to the AirBnb ESLint config includes a new rule that helps you avoid this particular pitfall by disallowing `bind` and arrow functions in JSX. We might consider trying to update the AirBnb config soon, as I think this would be useful, though there are some substantial changes (I did the update recently in a personal project and had a fair number of things to fix). We could also just [add the react/jsx-no-bind rule](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) to our config directly.

Sorry this is a pretty big PR... I think going through commit by commit might make it more manageable.

<s>TODO: I've still got to check that I solved the TransitionMotion exiting bug on the BGM side (I've already observed in the DOM that I've fixed it on the CGM side, though feel free to double-check me). I wanted to get this PR open ASAP!</s>

<s>Update: I definitely screwed up something in `SMBGDateLineAnimated`, so there will be at least one more small commit to fix.</s>

Update: I believed I've fixed the bug, and I pushed the commit.


jebeck:

src/components/trends/smbg/SMBGDateLineAnimated.js
```diff
@@ -96,67 +109,82 @@ class SMBGDateLineAnimated extends Component {
...
-
     const classes = cx({
       [styles.smbgPath]: true,
       [styles.highlightPath]: focusedDay === date,
     });
 
     // NOTE: This mapping is required due to the differing
     // expectations of TransitionMotion and d3 line
-    const mapObject = (obj, fn) => _.map(_.keys(obj), (key) => fn(obj[key], key, obj));
+    // const mapObject = (obj, fn) => _.map(_.keys(obj), (key) => fn(obj[key], key, obj));
```
krystophv: Commented code is commented. :)

jebeck: Ooops, thanks for pointing out. Deleted.

PR #51, opened 2017-02-02
CGM trends add delay before showing cbg traces

jebeck:
This adds some additional action creators to introduce a 250ms delay after hover on cbg slice segment before exposing the CGM traces that intersect with that segment. (The point of this is to make the hover interaction itself snappier since we don't clog up the DOM with inserting/removing potentially thousands of new nodes (individual cbg `<circle>`s) as soon as the mouse enters a segment.)

As mentioned I ended up not using redux saga for this to avoid complications with error handling/obscured stack traces in blip.

There are also a couple of small tweaks to animations as well as I was doing final testing and noticed small issues. Going through commit by commit may be easiest - each commit is quite clean and focused.


jebeck:

jebeck:
@hntrdglss not sure if you saw the ping for review on this one (and #52 as well)?

src/redux/actions/thunks.js
```diff
@@ -0,0 +1,40 @@
...
+
+import _ from 'lodash';
+
+import { focusTrendsCbgSlice, showCbgDateTraces } from './trends';
+
+export function delayShowCbgTracesOnFocus(userId, sliceData, slicePosition, focusedKeys) {
+  return (dispatch, getState) => {
+    dispatch(focusTrendsCbgSlice(userId, sliceData, slicePosition, focusedKeys));
+    setTimeout(() => {
+      const {
```
hntrdglss: I wonder if this could be clarified, through multiple variables? It's an impressive chunk, but with destructuring, `_.pick()`, and `_.get()`, there is a lot going on.

jebeck: Ah yeah sure it is quite a lot there - I'll break it out.

jebeck:
The build failed here after my change because it doesn't have 94b00b9646fbc8bc74a143006f38a1f2ce2198ab yet from #53. Since that'll come in with that PR and since this isn't going into master directly, just gonna leave it.

PR #55, opened 2017-02-09
handle display of out-of-range BGs on Trends

jebeck:
This change set is more complex than I anticipated. Would be interested in feedback to simplify it, but I'm not sure that's possible.

The root of the problem is that we were showing things like this in Trends, surfacing a fake 39 value (really Low Dexcom data):
screenshot 2017-02-08 19 56 05

Just as we do in e.g., the Daily view, this should show with 'Low' instead:
screenshot 2017-02-08 19 56 44
(We use the two-letter abbreviations 'Lo' and 'Hi' due to size constraints in the round tooltips on Daily view but will be phasing that out.)

The complication comes in with the fact that on Trends where we are showing *aggregate* data and not surfacing values from individual data points, we may have data from more than one device present (especially on the BGM side) and different devices have different out-of-range value thresholds.

So the general approach here is to:
1. find all the `bg/out-of-range` annotations in the data in view
2. determine the most conservative `low` and `high` thresholds from the annotation data
3. display 'Low' or 'High' as appropriate based on our calculated combined thresholds

@darinkrauss ran a query to confirm that in our current production data, *all* of our `bg/out-of-range` annotations (which have the form e.g., `{ code: 'bg/out-of-range', value: 'low', threshold: 40 }`) are in mg/dL units, so we only need to use the user's display units in the `displayBgValue` function. (That is, we don't need to worry about a step to convert all the thresholds to common units before determining the most conservative set... because they are all already in the same units.)

### DIFF
--- a/src/components/trends/smbg/FocusedSMBGPointLabel.js
+++ b/src/components/trends/smbg/FocusedSMBGPointLabel.js
@@ -38,6 +38,14 @@ const FocusedSMBGPointLabel = (props) => {
     return null;
   }
 
+  function getOutOfRangeThreshold(smbg) {
+    const outOfRangeAnnotation = _.find(
+      smbg.annotations || [], (annotation) => (annotation.code === 'bg/out-of-range')
+    );
+    return outOfRangeAnnotation ?
+      { [outOfRangeAnnotation.value]: outOfRangeAnnotation.threshold } : null;
+  }
+
   const {
     bgUnits,
     focusedPoint: { datum, position, allSmbgsOnDate, allPositions },
@@ -57,7 +65,11 @@ const FocusedSMBGPointLabel = (props) => {
   const pointTooltips = _.map(allSmbgsOnDate, (smbg, i) => (
     <Tooltip
       key={i}
-      content={<span className={styles.number}>{displayBgValue(smbg.value, bgUnits)}</span>}
+      content={
+        <span className={styles.number}>
+          {displayBgValue(smbg.value, bgUnits, getOutOfRangeThreshold(smbg))}
+        </span>
+      }
       position={allPositions[i]}
       side={'bottom'}
       tail={false}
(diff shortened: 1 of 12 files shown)


src/components/trends/smbg/FocusedSMBGPointLabel.js
```diff
@@ -57,7 +65,11 @@ const FocusedSMBGPointLabel = (props) => {
   const pointTooltips = _.map(allSmbgsOnDate, (smbg, i) => (
     <Tooltip
       key={i}
-      content={<span className={styles.number}>{displayBgValue(smbg.value, bgUnits)}</span>}
+      content={
+        <span className={styles.number}>
+          {displayBgValue(smbg.value, bgUnits, getOutOfRangeThreshold(smbg))}
```
jebeck: This is the place where I'm least sure of my approach... Because here we *are* dealing with a single data point, so we could shortcut the somewhat annoying need to extract and reshape the relevant threshold (if any) with `getOutOfRangeThreshold` by just doing something like:
```
const outOfRangeAnnotation = _.find(
  smbg.annotations || [], (annotation) => (annotation.code === 'bg/out-of-range')
);
...then inside the <span>...
{outOfRangeAnnotation ? outOfRangeAnnotation.value : displayBgValue(smbg.value, bgUnits)}
```

The problem I think is that the aggregate data and the individual datums really would each be best suited by a `displayBgValue` function with a different interface. Ideally the display function for the individual datum BG values could take the whole datum and extract the annotations itself... In fact, in older tideline code [the equivalent of the `displayBgValue` function](https://github.com/tidepool-org/tideline/blob/master/js/data/util/format.js#L26-L40) took the entire datum in its params for precisely this reason. So it's a tough problem I think...

jebeck:

src/components/trends/smbg/FocusedSMBGPointLabel.js
krystophv: You make a decent point and I think it's probably good to keep it in mind going forward (if we find ourselves desiring a differentiation between the formatting functions more often), but since this isn't common yet and is executed a relatively few number of times, I think it's fine to just roll with this solution for now.

PR #56, opened 2017-02-13
Settings printview

jh-bate:
@krystophv  see https://trello.com/c/CYlJtmZg and https://github.com/tidepool-org/blip/pull/387 for context


jh-bate:
[<ux:2 dev:8 tst:2> Device settings print view in Blip](https://trello.com/c/CYlJtmZg/87-ux-2-dev-8-tst-2-device-settings-print-view-in-blip)

src/components/settings/NonTandem.js
```diff
@@ -150,7 +164,6 @@ const NonTandem = (props) => {
         </div>
```
krystophv: I think we'll need a `{renderPrintNotes()}` for the non-active schedules as well.

src/components/settings/settings.css
```diff
@@ -89,3 +90,53 @@
...
+  .basalTable tbody tr:last-of-type {
+    font-weight: normal;
+  }
+
+  .basalHeaderBackground, .bolusHeaderBackground {
+    background-image: none;
+    color: var(--text-black);
+    font-weight: normal;
+    composes: largeSize from '../../styles/typography.css';
+  }
```
krystophv: I think could use `padding-left: 0px` here.

src/components/settings/NonTandem.css
```diff
@@ -70,3 +71,39 @@
...
+    display: flex;
+    flex-flow: row wrap;
+    justify-content: space-around;
+  }
+
+  .bolusSettingsContainer {
+    display: flex;
+    flex-flow: row wrap;
+    justify-content: space-around;
+  }
```
krystophv: I think this and the `.basalSettingsContainer` above could both use `page-break-inside: avoid`s.

krystophv:
This LGTM. The print views don't exactly match the design in Zeplin, so you might need to get a pass on those via a design review (I personally like the striped tables).

krystophv:
LGTM 👍

PR #70, opened 2017-04-29
datetime utils cleanup

jebeck:
As discussed @krystophv I would go commit by commit - each commit is nicely contained.

The number of tests dropped by a very small number (2, I think) just because in moving one function into `TrendsContainer` and testing it as part of an instance method I couldn't easily do as many test cases w/o more complex test setup (to change the instance props). The harder case (timezone-aware) is the one being tested, so I think it's fine.


jebeck:

PR #71, opened 2017-05-02
format utils cleanup

jebeck:
Again as planned @krystophv 

The last commit isn't the greatest because I re-alphabetized (for ease of comparing across source and test files to make sure everything's covered), but otherwise commit-by-commit would be a good strategy for review I'd imagine.

There is [a blip PR](https://github.com/tidepool-org/blip/pull/413) that will depend on this since I standardized on passing through entire `bgPrefs` to the `TrendsContainer` and Trends tooltip components.


jebeck:

PR #72, opened 2017-05-02
get rid of src/containers/ in favor of everything in src/components/

jebeck:
I based this off #71 for expediency, but I'd review that first @krystophv to keep the diff sane. I've done a fair amount of running this locally to prove to myself that the reorganization hasn't broken anything, so I'm fairly confident we're good to go here.


jebeck:

PR #73, opened 2017-05-03
relax node engine to avoid Yarn install failures

jebeck:
See [this Travis build for blip](https://travis-ci.org/tidepool-org/blip/jobs/228490914) where Yarn failed out due to viz specifying a node engine of 6.x.

This is a dependency for [blip #414](https://github.com/tidepool-org/blip/pull/414).

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "url": "git+https://github.com/tidepool-org/viz.git"
   },
   "engines": {
-    "node": "6.x",
+    "node": ">=6.x",
     "npm": "3.x"
   },
   "license": "BSD-2-Clause",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "0.7.19",
+  "version": "0.7.20",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "0.7.20",
+  "version": "0.8.1",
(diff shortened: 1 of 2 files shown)


jebeck:
Perhaps just double-check me on the new commit 8772219dd9593052adc8868c7c3f195f118e611e @krystophv?

package.json
```diff
@@ -9,11 +9,12 @@
   "scripts": {
     "apidocs": "./node_modules/.bin/jsdoc2md",
     "browser-tests": "NODE_ENV=test ./node_modules/.bin/karma start --browsers Chrome",
-    "build": "NODE_ENV=production npm test && npm run clean && ./node_modules/.bin/webpack --config package.config.js --optimize-minimize",
+    "build": "NODE_ENV=production npm run clean && ./node_modules/.bin/webpack --config package.config.js --optimize-minimize",
     "build-docs": "./update-gh-pages.sh",
     "clean": "./node_modules/.bin/rimraf ./dist/*",
     "lint": "./node_modules/.bin/eslint .storybook/ src/ stories/ test/ *.js",
-    "prepublish": "npm run build",
+    "prepare": "npm run build",
```
jebeck: Ok, there is one weird thing about this config. I chose to run the build only on `prepare` because when you `npm link` it _is_ nice to have certainty that a build will exist in dist/ (e.g., if you forget to `npm start`, which I have done). But `prepare` always runs before `prepublishOnly`, so that means we build, then delete everything, reinstall, and build again. I'm open to removing `prepare` altogether...what do you think @krystophv?

krystophv: I think it's probably alright to have it the way it is here. It is a little awkward, but it only happens on `publish`, which is not terribly often and it does solve the issue with having it test and lint on install.

PR #74, opened 2017-05-12
render target agnostic bolus rendering

jebeck:
This is a port of tideline's bolus rendering (pretty much [this file](https://github.com/tidepool-org/tideline/blob/master/js/plot/util/drawbolus.js) and the filtering in [this one](https://github.com/tidepool-org/tideline/blob/master/js/plot/wizard.js)).

Still to come: the carb circles on calculator-accompanied ("wizard") insulin events. I thought this was enough to review for now.

The most important things to review here are the additions to the bolus utilities and the actual path generation. There is also a React component (& tests) to render bolus paths in SVG on the web for Storybook purposes (at present). The commits are fairly logically grouped sets of changes, but some may still be large sets of changes.

LMK if you have any questions of would like a walk-through @krystophv

### DIFF
--- a/src/utils/bolus.js
+++ b/src/utils/bolus.js
@@ -129,6 +129,43 @@ export function getDelivered(insulinEvent) {
 }
 
 /**
+ * getDuration
+ * @param {Object} insulinEvent - a Tidepool bolus or wizard object
+ *
+ * @return {Number} duration value in milliseconds
+ */
+export function getDuration(insulinEvent) {
+  let bolus = insulinEvent;
+  if (_.get(insulinEvent, 'type') === 'wizard') {
+    bolus = getBolusFromInsulinEvent(insulinEvent);
+  }
+  // don't want truthiness here because want to return duration
+  // from a bolus interrupted immediately (duration = 0)
+  if (bolus.duration == null) {
+    return NaN;
+  }
+  return bolus.duration;
+}
+
+/**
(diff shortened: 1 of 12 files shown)


jebeck:

src/utils/bolus.js
```diff
@@ -129,6 +106,66 @@ export function getRecommended(insulinEvent) {
...
+ * @return {Number} duration value in milliseconds
+ */
+export function getDuration(insulinEvent) {
+  let bolus = insulinEvent;
+  if (_.get(insulinEvent, 'type') === 'wizard') {
+    bolus = getBolusFromInsulinEvent(insulinEvent);
+  }
+  // don't want truthiness here because want to return duration
+  // from a bolus interrupted immediately (duration = 0)
+  if (bolus.duration == null) {
```
jebeck: The check here is actually mostly just against `undefined`...but I used `== null` because that handles both `undefined` and `null` through coercion.

I definitely see your point about explicitness. After looking the Lodash 3.x docs again, perhaps `if (!_.inRange(bolus.duration, Infinity))` would be most explicit? That requires both finiteness and `>= 0`, which is pretty much exactly the requirements of the value (for the purposes of rendering).

krystophv: Yeah, that seems to nail down the requirements for the value pretty well. LGTM.

PR #76, opened 2017-05-17
bolus rendering tweaks + more ddata utils adds

jebeck:
There are two things here: some small tweaks to the bolus rendering for greater render-target-agnostic behavior (after trying things out in print view)... The key here is to define all paths explicitly as shapes even if just drawing a line because of the variability across render targets for how stroke is positioned on a path. (Yes, it varies, sadly.)

Then I pulled over more of the utilities I've had to add in the print view work, including a backwards-compatible change to the existing `classifyBgValue`.

Would recommend going through this PR commit by commit @krystophv

### DIFF
--- a/src/components/common/data/Bolus.css
+++ b/src/components/common/data/Bolus.css
@@ -26,7 +26,7 @@
 }
 
 .extendedExpectationPath {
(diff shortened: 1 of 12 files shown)


jebeck:

PR #77, opened 2017-05-17
agnostic basal rendering

jebeck:
All right, I lied about getting the docs ready for this PR because I would like to merge this so that I can consolidate my branches - it's getting a little confusing trying to separate out the utils work & render target agnostic rendering bits and then having to reintegrate them into my daily print view progress. So docs will come as part of a general effort to prepare the daily print view work for handoff.

Without docs, I imagine this is quite dense, so LMK if you want a bit of a walk-through before reviewing @krystophv 

Do **NOT** review this one commit-by-commit; I changed my approach pretty radically and redid a bunch of the rendering halfway through...

PR #78, opened 2017-05-18
additional utils for daily print view stats

jebeck:
Some pretty simple additions here. Branched off #77 just to keep my workflow simpler as the time crunch is getting more real...


jebeck:

PR #79, opened 2017-05-18
render carb circles & labels for wizards

jebeck:
Another small addition building on the last PR (#78).

Note that I've chosen to add the rendering for the web SVG target independently of the print view here. While it would be possible to do the circles with a path generator, it's not possible to do the text label that way, so some code would have to be reimplemented across print view and web SVG rendering anyway, and circles are easy, so might as well just reimplement both rather than have a confusing (IMO) mix of circle rendered via the path generator and label rendered in the native code.


jebeck:

PR #84, opened 2017-06-17
Daily Print View

cbwebdevelopment:
This allows the daily chart data for the most recent 6 days to be printed from Tidepool Web when viewing the Daily view.

Goes hand-in-hand with corresponding `clintb/daily-print-view.web-worker` branch on blip.

Details at https://trello.com/c/tjD655oz


cbwebdevelopment:
[<L> [blip] Daily print view first production release](https://trello.com/c/tjD655oz/182-l-blip-daily-print-view-first-production-release)

test/modules/print/DailyPrintView.test.js
```diff
@@ -14,3 +14,348 @@
...
+          topEdge: opts.margins.top,
+          width: Renderer.rightEdge - (opts.margins.left +
+          (opts.summaryWidthAsPercentage + Renderer.gapBtwnSummaryAndChartAsPercentage) *
+          Renderer.width),
+        } },
+      ];
+
+      _.each(requiredProps, item => {
+        expect(Renderer[item.prop]).to.be.a(item.type);
+        item.value && expect(Renderer[item.prop]).to.eql(item.value);
```
krystophv: Won't this `item.value` return false for numeric `0` and short circuit the `expect`?

src/utils/print/data.js
```diff
@@ -57,6 +57,29 @@ function stripDatum(d) {
   ));
 }
 
+// eslint-disable-next-line require-jsdoc
```
cbwebdevelopment: Not sure... These were put in by Jana.  I'm thinking it had to do with not wanting to a have docs auto-generated for them when `npm run apidocs` was run.

They were initially internal functions inside of `selectDailyViewData`, and I'm assuming she didn't want to advertise their use outside of it.

I ended up moving them outside to allow me to export them for proper unit testing.

I don't see any harm in documenting them, though. so I'll just go ahead and do that.

test/modules/print/DailyPrintView.test.js
cbwebdevelopment: Yep, you're right!  Again, good catch here!

cbwebdevelopment:
@krystophv I've addressed the issues you brought up

PR #88, opened 2017-09-18
Daily View Print View cannot show bolus wizard carb counts >99g (only shows two digits)

cbwebdevelopment:
Just had to increase the width of the printed carb count to allow displaying any reasonable amount.

Goes hand in hand with PR in `blip`:
https://github.com/tidepool-org/blip/pull/444


cbwebdevelopment:
[Daily View Print View cannot show bolus wizard carb counts >99g (only shows two digits)](https://trello.com/c/t777c7sr/249-daily-view-print-view-cannot-show-bolus-wizard-carb-counts-99g-only-shows-two-digits)

PR #89, opened 2017-09-29
Fix for displaying inaccurate 'High' and 'Low' smbg values

cbwebdevelopment:
For CBG values, we show a HI/LO tooltip when the value is out of range. For SMBG values, we show value +1/-1 the threshold, e.g. when the threshold is 600 the value would be 601, and an annotation to indicate that the actual value can be higher or lower. 

We want CBG and SMBG values to have the same behaviour, ie., both should show 'High/Low' text (CBG text to be truncated to 'Hi/Lo' to fit into small tooltip circle.

Full details on trello: https://trello.com/c/p7L3RoNn

Goes hand-in-hand with https://github.com/tidepool-org/tideline/pull/321

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "0.9.0",
+  "version": "0.9.2",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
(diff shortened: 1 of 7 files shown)


cbwebdevelopment:
[HI/LO SMBG shows actual value instead](https://trello.com/c/p7L3RoNn/260-hi-lo-smbg-shows-actual-value-instead)

PR #90, opened 2017-10-25
bugfix for Redux state mutation violation (crashes app in dev mode)

jebeck:
Prior to this fix, hovering on a BGM in default dev tools on mode (with the Redux state mutation tracker) would crash the app when hovering on a BGM in the BGM version of Trends view:
screenshot 2017-10-25 09 24 37
screenshot 2017-10-25 09 24 42
screenshot 2017-10-25 09 24 47

The change I've made here is to make sure to clone the `allSmbgsOnDate` and `allPositions` arrays (via `.slice()`) in `<FocusedSMBGPointLabel/>` prior to mutating them.

Just a small little fix 🙂

### DIFF
--- a/src/components/trends/smbg/FocusedSMBGPointLabel.js
+++ b/src/components/trends/smbg/FocusedSMBGPointLabel.js
@@ -57,9 +57,9 @@ const FocusedSMBGPointLabel = (props) => {
   let smbgsOnDate = allSmbgsOnDate.slice();
   let positions = allPositions.slice();
   if (!lines) {
-    const focusedPointIndex = _.indexOf(allSmbgsOnDate, datum);
-    smbgsOnDate = _.pullAt(smbgsOnDate, focusedPointIndex);
-    positions = _.pullAt(positions, focusedPointIndex);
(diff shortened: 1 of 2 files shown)


jebeck:
Ooops, I've got a test or two to fix I see. I'll do that in a bit!

jebeck:
All green now 😁  The place where I change the assertions in tests from 4 to 3 makes sense to me—there are only three smbgs in the fixture, so a maximum of three tooltips should appear. The old `_.indexOf` that I changed to `_.findIndex` wasn't finding anything in the test fixtures, resulting in nothing dropped via the `_.pullAt`s, I think.

LMK if you have any questions! I thought this was a good small bug to start with since it mainly affects dev mode (though it also left the door open for other bugs, since mutating things in your Redux store outside of reducers is a bad idea...so a definite code quality improvement...just a teensy one 😉 ).

cbwebdevelopment:
[Bugfix for Redux state mutation violation (crashes app in dev mode)](https://trello.com/c/6gwIwBIF/290-bugfix-for-redux-state-mutation-violation-crashes-app-in-dev-mode)

PR #93, opened 2017-11-24
Only volume-mount the bare minimum of files docker image.

cbwebdevelopment:
Goes hand-in-hand with:
https://github.com/tidepool-org/tools/pull/64
https://github.com/tidepool-org/blip/pull/451

Changes to Dockerfile to allow us to not volume-mount the `node_modules` and `dist` directories to improve local performance.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,8 +19,6 @@ COPY . /app
 RUN yarn install
 
 VOLUME /app
-VOLUME /app/node_modules
-VOLUME /app/dist
 
 EXPOSE 8081 8082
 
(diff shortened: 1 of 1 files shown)


cbwebdevelopment:
[Docker performance improvements for 'blip' and 'viz' services](https://trello.com/c/gJMqJ9K6/287-docker-performance-improvements-for-blip-and-viz-services)

Dockerfile
```diff
@@ -8,15 +8,20 @@ RUN apk add --no-cache fontconfig \
...
 
+USER node
+
 COPY . /app
 
-USER node
+RUN yarn install
+
+VOLUME /app
+VOLUME /app/node_modules
```
pazaan: Does exposing node_modules make sense? Compiled modules wouldn't work across different OSes.

cbwebdevelopment: You're right, there's no good reason for this.  I've removed it.

PR #94, opened 2017-12-01
Daily y-scale fix for data sets with no values above target range

cbwebdevelopment:
Data with no above-target values causes Y scale on daily view to be off (the very high boundary is rendered outside of the BG chart area)

See https://trello.com/c/1r7IxaiA for details.

Related PRs
https://github.com/tidepool-org/blip/pull/455
https://github.com/tidepool-org/tideline/pull/324


cbwebdevelopment:
[Data with no above-target values causes Y scale on daily view to be off.](https://trello.com/c/1r7IxaiA/289-data-with-no-above-target-values-causes-y-scale-on-daily-view-to-be-off)

src/modules/print/DailyPrintView.js
```diff
@@ -742,7 +749,10 @@ class DailyPrintView {
       align: 'right',
       width: this.chartArea.leftEdge - this.summaryArea.rightEdge - 3,
     };
-    _.each(this.bgBounds, (bound) => {
+
+    const renderedBounds = _.filter(this.bgBounds, bound => bound <= this.bgScaleYLimit);
```
krystophv: The function notation here gave me pause. `=> var <=` looked pretty strange at a glance.

cbwebdevelopment: I added some parentheses around the returned comparison.

PR #95, opened 2017-12-19
Bolus tooltips

krystophv:
Adds a new bolus tooltip which will display all relevant information concerning a bolus.

See also:
blip: https://github.com/tidepool-org/blip/pull/457
tideline: https://github.com/tidepool-org/tideline/pull/325

### DIFF
--- a/src/components/daily/bolustooltip/BolusTooltip.js
+++ b/src/components/daily/bolustooltip/BolusTooltip.js
@@ -25,6 +25,17 @@ import colors from '../../../styles/colors.css';
 import styles from './BolusTooltip.css';
 
 class BolusTooltip extends PureComponent {
+  formatInsulin(qty) {
+    let decimalLength;
+    const qtyString = qty.toString();
+    if (qtyString.indexOf('.') !== -1 && qtyString.split('.')[1].length === 2) {
+      decimalLength = 2;
+    } else {
+      decimalLength = 1;
+    }
+    return formatDecimalNumber(qty, decimalLength);
+  }
+
   getTarget() {
     const wizardTarget = _.get(this.props.bolus, 'bgTarget');
     const target = _.get(wizardTarget, 'target', null);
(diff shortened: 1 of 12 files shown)


krystophv:
[<XL> [data viz] [daily] Differentiate bolus from carbs vs. bolus from correction](https://trello.com/c/vFjtdPcF/141-xl-data-viz-daily-differentiate-bolus-from-carbs-vs-bolus-from-correction)

src/components/daily/bolustooltip/BolusTooltip.js
```diff
@@ -0,0 +1,331 @@
...
+  }).isRequired,
+  timePrefs: PropTypes.object.isRequired,
+};
+
+BolusTooltip.defaultProps = {
+  tail: true,
+  side: 'right',
+  tailWidth: 9,
+  tailHeight: 17,
+  tailColor: colors.bolus,
```
cbwebdevelopment: Missing declaration for `tailColor` in `BolusTooltips.propTypes`

stories/components/daily/BolusTooltip.js
```diff
@@ -0,0 +1,479 @@
+import React from 'react';
+
+import { storiesOf } from '@kadira/storybook';
+
+import BolusTooltip from '../../../src/components/daily/bolustooltip/BolusTooltip';
+
+// bolus' lifted from bolus utils tests
```
cbwebdevelopment: Just wondering what your thoughts are on whether or not it's better to move this into a shared data fixture.  I'm honestly not sure which I prefer.  I like code reuse, but it's also nice to have the context in place when it's just being used in tests/stories.

krystophv: Yeah, I had considered moving it into a shared fixture, but the value in being able to just scroll to/jump to the definition in the same file outweighed the DRY aspect for me, but that may have just been because of how often I was referencing it during development. If we find ourselves using the fixture anywhere else, I'd strongly recommend collapsing it.

PR #117, opened 2018-07-27
Add thousandths precision to insulin formatting

krystophv:
Some pumps allow for boluses at 0.025 unit intervals and we shouldn't be rounding them if they're provided in that level of precision.

blip PR: https://github.com/tidepool-org/blip/pull/502

Trello: https://trello.com/c/9ET8Bmh5/146-setting-6-series-bolus-increment-to-025-causes-loss-of-precision-due-to-rounding

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "0.11.2",
+  "version": "0.11.3-bolus-rounding.1",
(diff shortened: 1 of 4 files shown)


krystophv:
[Setting 6-series bolus increment to .025 causes loss of precision due to rounding.](https://trello.com/c/9ET8Bmh5/146-setting-6-series-bolus-increment-to-025-causes-loss-of-precision-due-to-rounding)

PR #120, opened 2018-11-26
Aggregated stats widgets

cbwebdevelopment:
See https://trello.com/c/suWzyusJ for details.

In addition to the new stat widgets components and data processing, this PR includes a major sweep of NPM package updates

The npm dependancies have been updated as much as as possible. React has only been bumped to latest 15.x.

The dev dependancies are all updated to use the latest version (i.e. webpack, babel, storybook, test-runners, etc).

Note: there are currently a couple small dev items that are still in progress, for which I will ping for review as they become ready:

1. There is a new type of stat required for the `Avg. daily dose / weight` stat, which requires some refactoring to the `Stat.js` component.  I'll complete the tests in `Stat.test.js` once the component development is finalized
2. I need to ensure the strings in 'utils/stat.js' are translateable.

Related PRs
---
tidepool-org/tideline#357
tidepool-org/blip#513

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -22,6 +22,7 @@
     "import/prefer-default-export": 0,
     "jsx-a11y/alt-text": 0,
     "jsx-a11y/click-events-have-key-events": 0,
+    "jsx-a11y/label-has-for": 0,
     "jsx-a11y/mouse-events-have-key-events": 0,
     "jsx-a11y/no-noninteractive-element-interactions": 0,
     "jsx-a11y/no-static-element-interactions": 0,
--- a/src/components/settings/settings.css
+++ b/src/components/settings/settings.css
@@ -51,7 +51,6 @@
 
 .settingsTable {
   composes: defaultSize mediumContrastText from '../../styles/typography.css';
-  /* min-width: 250px; */
   width: 100%;
   border-spacing: 1px 0px;
   border-collapse: separate;
--- a/src/components/trends/cbg/CBGMedianAnimated.js
+++ b/src/components/trends/cbg/CBGMedianAnimated.js
@@ -111,7 +111,7 @@ export class CBGMedianAnimated extends PureComponent {
 
     const strokeWidth = sliceWidth / 8;
     const medianWidth = sliceWidth - strokeWidth;
-    const medianHeight = medianWidth * 0.758;
+    const medianHeight = medianWidth * 0.75;
     const binLeftX = xScale(datum.msX) - medianWidth / 2 + strokeWidth / 2;
     const width = medianWidth - strokeWidth;
 
--- a/test/components/common/stat/Stat.test.js
+++ b/test/components/common/stat/Stat.test.js
@@ -3,7 +3,7 @@ import { shallow } from 'enzyme';
 
 import Stat from '../../../../src/components/common/stat/Stat';
 
-describe.only('Stat', () => {
+describe('Stat', () => {
   let wrapper;
 
   const defaultProps = {
--- a/test/components/common/stat/Stat.test.js
+++ b/test/components/common/stat/Stat.test.js
@@ -1,12 +1,30 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 
+import { formatClassesAsSelector } from '../../../helpers/cssmodules';
 import Stat from '../../../../src/components/common/stat/Stat';
+import styles from '../../../../src/components/common/stat/Stat.css';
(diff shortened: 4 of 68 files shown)


cbwebdevelopment:
@krystophv I've finished up with the remaining work on this.  This includes the TDD/Weight stat of type 'input', and a corresponding `InputGroup` component.

I've also got all the unit tests updated/completed, including the following new ones (since the initial PR):
BgBar.test.js
BgBarLabel.test.js
HoverBar.test.js
HoverBarLabel.test.js
InputGroup.test.js
Stat.test.js

The new commits start at `5b37c2d0`

.eslintrc
```diff
@@ -3,9 +3,60 @@
   "parser": "babel-eslint",
   "plugins": ["lodash"],
   "rules": {
-    "lodash/prefer-lodash-method": "error"
-  },
-  "settings": {
-    "lodash": 3
+    "arrow-parens": 0,
```
krystophv: This is a whole lot of rule exceptions. Did these come in as a prescripted set from somewhere or mostly preferential one-offs?

test/components/common/stat/Stat.test.js
```diff
@@ -0,0 +1,2545 @@
...
+        });
+
+        expect(instance.formatDatum({
+          value: 100,
+        }, statFormats.bgValue)).to.include({
+          id: 'target',
+          value: '100',
+        });
+
+        // Using 3-way classification, so both `high
```
krystophv: Should be 2 decimal places?

krystophv: Should be 2 decimal places?

src/components/settings/settings.css
```diff
@@ -51,7 +51,7 @@
 
 .settingsTable {
   composes: defaultSize mediumContrastText from '../../styles/typography.css';
-  min-width: 250px;
+  /* min-width: 250px; */
```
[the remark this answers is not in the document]
cbwebdevelopment: Yep.  Will do

src/components/trends/cbg/CBGMedianAnimated.js
```diff
@@ -94,26 +92,28 @@ export class CBGMedianAnimated extends PureComponent {
...
       cx({
         [styles.median]: true,
         [styles.transparent]: true,
       });
 
-    const binLeftX = xScale(datum.msX) - medianWidth / 2 + styles.stroke / 2;
-    const width = medianWidth - styles.stroke;
+    const strokeWidth = sliceWidth / 8;
+    const medianWidth = sliceWidth - strokeWidth;
+    const medianHeight = medianWidth * 0.758;
```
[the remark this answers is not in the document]
cbwebdevelopment: Yeah, I'll have to look at that one (was quite a while ago now).  I remember having a calculator when trying to find a value that worked well.  I'll figure out what I did, and put a useful comment in here.

.eslintrc
cbwebdevelopment: When I upgraded to the latest versions of the various eslint packages we had been using, it started flagging a tonne of new issues.  

I decided that it would be better to simply add rule exceptions for most of the new issues, rather than fix all occurrences to satisfy the new lint rules, which would have amounted to a massive amount of changes to the codebase.

I do think that we should likely tackle this at some point, but I feel like we should perhaps figure out exactly what rules/configs we want to hold to, and likely consolidate `viz` and `mobile` to use the same configs so we have some consistency moving forward.

What do you think, Chris?

src/components/trends/cbg/CBGMedianAnimated.js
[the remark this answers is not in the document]
cbwebdevelopment: Ok, so I have no idea where this came from.  The original, hard-coded height/width would have resulted in a magic number of `.714`.  I'm thinking now that I'd intended to make it a 3/4 ratio, and accidentally fat-fingered an 8 on the end.  I'll put it to `.75`

.eslintrc
krystophv: That seems reasonable. I do think it simplifies a lot of things when we're able to use these sorts of "standards" (eg airbnb eslint config) more out of the box than making a lot of edits, but I do agree that applying the fixes directly to this PR would have been foolhardy. I think that we should look at a consistent eslint across our codebases as much as possible (`viz`, `mobile`, `uploader` and `blip` should all be reasonable to keep together I think), and that it should probably happen in a concerted effort.

PR #121, opened 2018-12-05
Auto suspend markers

krystophv:
blip: https://github.com/tidepool-org/blip/pull/514
tideline: https://github.com/tidepool-org/tideline/pull/358
trello: https://trello.com/c/wZk0cMNU

### DIFF
--- a/src/components/common/data/Suspend.js
+++ b/src/components/common/data/Suspend.js
@@ -90,7 +90,7 @@ Suspend.propTypes = {
       duration: PropTypes.number.isRequired,
       utc: PropTypes.number.isRequired,
       id: PropTypes.string.isRequired,
-    }).isRequired
+    }).isRequired,
   ).isRequired,
   flushBottomOffset: PropTypes.number.isRequired,
   xScale: PropTypes.func.isRequired,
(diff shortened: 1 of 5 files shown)


krystophv:
[Distinguish between auto and manual suspend for PLGS (like Basal-IQ)](https://trello.com/c/wZk0cMNU/488-distinguish-between-auto-and-manual-suspend-for-plgs-like-basal-iq)

src/components/common/data/Suspend.js
```diff
@@ -0,0 +1,100 @@
...
+
+Suspend.propTypes = {
+  suspends: PropTypes.arrayOf(
+    PropTypes.shape({
+      type: PropTypes.oneOf(['deviceEvent']).isRequired,
+      subType: PropTypes.oneOf(['suspend']).isRequired,
+      duration: PropTypes.number.isRequired,
+      utc: PropTypes.number.isRequired,
+      id: PropTypes.string.isRequired,
+    }).isRequired
```
cbwebdevelopment: I'll suggest a trailing comma here, if only because with the package updates in the agg stat work looming, this will be caught by the linter.  Unless we prefer to not have the linter catch these.

krystophv: I'm totally down for having the linter catch this. It's just that old (like IE 5/6/7 would die on hanging commas for arrays old) habits die hard.

PR #129, opened 2019-02-19
Carb rendering

krystophv:
Adds a FoodTooltip which largely mimics the existing tooltip components, if just a little simplified for now as we _only_ have/are concerned with displaying carbohydrate data from an otherwise potentially very complex datatype (`food`). Also modifies `getCarbsData` to take the new `food` datatype into account.

Pairs well with:
tideline: https://github.com/tidepool-org/tideline/pull/365
blip: https://github.com/tidepool-org/blip/pull/537

Trello: https://trello.com/c/0WUNaQS6

### DIFF
--- a/data/types.js
+++ b/data/types.js
@@ -334,11 +334,31 @@ export class Wizard extends Common {
   }
 }
 
+export class Food extends Common {
+  constructor(opts = {}) {
+    super(opts);
+
+    _.defaults(opts, {
+      deviceTime: this.makeDeviceTime(),
+    });
+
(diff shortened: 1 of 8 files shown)


krystophv:
[Display Carbs from platform, which are not associated with Bolus Calculator (i.e. wizard) event in Tidepool Web](https://trello.com/c/0WUNaQS6/532-display-carbs-from-platform-which-are-not-associated-with-bolus-calculator-ie-wizard-event-in-tidepool-web)

PR #132, opened 2019-02-25
updated color of std deviation stat to match others in aggregated stats

thacherT1D:


thacherT1D:
testing -- changes duplicated by stats-std-dev-color

PR #133, opened 2019-02-28
Persist Average Daily Dose input between data views

cbwebdevelopment:
See https://trello.com/c/8w4zfRPq for details.

Goes with tidepool-org/blip#542

### DIFF
--- a/src/components/common/stat/Stat.js
+++ b/src/components/common/stat/Stat.js
@@ -71,6 +71,7 @@ class Stat extends PureComponent {
     isOpened: PropTypes.bool,
     legend: PropTypes.bool,
     muteOthersOnHover: PropTypes.bool,
+    onInputChange: PropTypes.func,
     reverseLegendOrder: PropTypes.bool,
     title: PropTypes.string.isRequired,
     type: PropTypes.oneOf(_.keys(statTypes)),
@@ -886,7 +887,7 @@ class Stat extends PureComponent {
     event.persist();
     this.setState(() => ({
       inputValue: event.target.value,
-    }));
+    }), this.propagateInputChange);
   };
 
   handleSuffixChange = value => {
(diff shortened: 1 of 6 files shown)


cbwebdevelopment:
[Regression: User-entered data for the new Avg. Daily Insulin widget does not persist](https://trello.com/c/8w4zfRPq/578-regression-user-entered-data-for-the-new-avg-daily-insulin-widget-does-not-persist)

test/components/common/stat/Stat.test.js
```diff
@@ -570,6 +570,14 @@ describe('Stat', () => {
       expect(inputGroup().props().onChange).to.be.a.function;
       expect(inputGroup().props().onSuffixChange).to.be.a.function;
     });
+
+    it('should should call the `onInputChange` Stat prop when provided on input change', () => {
+
```
krystophv: Bit of an echo in here. I think these tests might have been glanced over after being stubbed.

test/components/common/stat/Stat.test.js
```diff
@@ -2564,5 +2593,40 @@ describe('Stat', () => {
 
       expect(wrapper.state().inputSuffix.value).to.equal(600);
     });
+
+    it('should call `propagateInputChange` after setting state', () => {
```
krystophv: This is the same description as one added just above. Both are testing various state setting, but maybe the descriptions could be differentiated just a little?

src/components/common/stat/Stat.js
```diff
@@ -886,15 +887,21 @@ class Stat extends PureComponent {
...
     this.setState((state) => ({
       inputSuffix: _.assign({}, state.inputSuffix, {
         value,
       }),
-    }));
+    }), this.propagateInputChange);
+  };
+
+  propagateInputChange = () => {
+    if (typeof this.props.onInputChange === 'function') {
```
krystophv: Not a blocker, but should we consider using `_.isFunction` here as it's mildly more robust?

cbwebdevelopment: 👍

test/components/common/stat/Stat.test.js
cbwebdevelopment: I filled in the test stub for assigning default input and suffix values from props, and deleted the test stub for calling onInputChange, which is actually handled in the `propagateInputChange` tests, and the `handleInputChange` and `handleSuffixChange` tests that assert that it's called.

test/components/common/stat/Stat.test.js
cbwebdevelopment: I've differentiated them.

cbwebdevelopment:
@krystophv Ping for review of requested changes.

PR #138, opened 2019-03-11
Tidepool web crashes when rendering SD stat widget when SD > mean

cbwebdevelopment:
See https://trello.com/c/YcHZj4dI for details

Goes with:
tidepool-org/blip#549

Fix is twofold:

1. The actual error was resulting from the `classifyBgValue` utility function being called with a negative value during the BgBar component render. This PR ensures that only a positive, numeric value is passed in, which avoids the thrown error.
2. In cases where there are extremely wide deviation, it was possible that the high and low "bars" on the bg scale would actually render beyond the scale limits.  The rendered bar x positions are now constrained to the scale limits.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "1.1.0",
+  "version": "1.2.0-negative-sd-range-fix.1",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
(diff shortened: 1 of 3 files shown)


cbwebdevelopment:
[Tidepool web crashes when rendering SD stat widget when SD > mean](https://trello.com/c/YcHZj4dI/588-tidepool-web-crashes-when-rendering-sd-stat-widget-when-sd-mean)

PR #146, opened 2019-04-24
Build docker images in travis.

derrickburns:
This builds Docker images in Travis for each push that passes unit tests.

### DIFF
--- a/artifact.sh
+++ b/artifact.sh
@@ -1,6 +1,6 @@
 #!/bin/sh -e
 
-wget -q -O artifact_none.sh 'https://raw.githubusercontent.com/tidepool-org/tools/ci-changes/artifact/artifact.sh'
+wget -q -O artifact_none.sh 'https://raw.githubusercontent.com/tidepool-org/tools/master/artifact/artifact.sh'
 chmod +x artifact_none.sh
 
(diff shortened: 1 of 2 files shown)


artifact.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/sh -e
+
+wget -q -O artifact_none.sh 'https://raw.githubusercontent.com/tidepool-org/tools/ci-changes/artifact/artifact.sh'
```
pazaan: Why `artifact_none`? Isn't it producing a docker image (which is an artifact)?

derrickburns: I will change the names in the next pass.  The name it cannot be is artifact.sh!

pazaan: Ah, of course.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "1.3.0-weekly-to-bg-log.2",
+  "version": "1.3.0",
```
pazaan: Should this have a different version? @cbwebdevelopment

derrickburns: Good catch. I rebased on top of develop.  So, Clint's commits are not part of this now.

PR #149, opened 2019-05-08
Use artifact.sh

derrickburns:


artifact.sh
```diff
@@ -3,4 +3,4 @@
 wget -q -O artifact_none.sh 'https://raw.githubusercontent.com/tidepool-org/tools/master/artifact/artifact.sh'
 chmod +x artifact_none.sh
 
-./artifact_none.sh
+./artifact_none.sh none
```
pazaan: That `none` naming again ;) Maybe `docker` instead of `none`?

PR #167, opened 2019-10-09
Bg values outside chart limits [WEB-198]

thacherT1D:


src/modules/print/DailyPrintView.js
```diff
@@ -247,7 +247,7 @@ class DailyPrintView extends PrintView {
     this.data.bgRange[1] = _.max([this.data.bgRange[1], this.bgBounds.targetUpperBound]);
 
     // Calculate the maximum BG yScale value
-    this.bgScaleYLimit = _.min([this.data.bgRange[1], this.bgBounds.veryHighThreshold]);
+    this.bgScaleYLimit = _.max([this.data.bgRange[1], this.bgBounds.veryHighThreshold]);
```
cbwebdevelopment: Just wanted to call out that this now differs from the functionality on the web daily view to make sure the difference in handling out-of-scale values is intentional.

On the web, we clamp the bg values at 400 (22.5 mmol/L).  Here, we're adjusting the scale to whatever the highest BG value is.

To this point, the instruction has always been to mimic the web view logic as closely as possible in the print views, so I wanted to make sure product is OK with the change.

test/modules/print/DailyPrintView.test.js
```diff
@@ -230,14 +230,14 @@ describe('DailyPrintView', () => {
       Renderer.data.bgRange[1] = 100;
       Renderer.makeScales(Renderer.chartsByDate[sampleDate]);
 
-      expect(Renderer.bgScaleYLimit).to.equal(180);
+      expect(Renderer.bgScaleYLimit).to.equal(300);
```
cbwebdevelopment: The test description should be changed if we are now setting the `bgScaleYLimit` to a minimum of the very high threshold.

test/modules/print/DailyPrintView.test.js
```diff
@@ -230,14 +230,14 @@ describe('DailyPrintView', () => {
...
     });
 
     it('should set the bgScaleYLimit to a maximum of the BG very high threshold', () => {
       Renderer.data.bgRange[1] = 600;
       Renderer.makeScales(Renderer.chartsByDate[sampleDate]);
 
-      expect(Renderer.bgScaleYLimit).to.equal(300);
+      expect(Renderer.bgScaleYLimit).to.equal(600);
```
cbwebdevelopment: The test description should be changed if we are no longer setting a max for the `bgScaleYLimit`

src/modules/print/DailyPrintView.js
cbwebdevelopment: I should also note that the current functionality doesn't mimic the web view either, in that it doesn't clamp any values.

It may be that the change you've requested is indeed the proper approach.  I just want to be sure.

thacherT1D:
Closed due to updated Done Criteria, a different engineering solution is required to address the new Done Criteria

PR #186, opened 2020-02-07
[WEB-706] react upgrade

krystophv:
[WEB-706] - upgrading react to latest version

accompanied by:
https://github.com/tidepool-org/blip/pull/650
https://github.com/tidepool-org/tideline/pull/390

[WEB-706]: https://tidepool.atlassian.net/browse/WEB-706


src/components/common/data/Basal.js
```diff
@@ -1,22 +1,7 @@
...
- *
- * This program is distributed in the hope that it will be useful, but WITHOUT
- * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
- * FOR A PARTICULAR PURPOSE. See the License for more details.
- *
- * You should have received a copy of the License along with this program; if
- * not, you can obtain one from Tidepool Project at tidepool.org.
- * == BSD2 LICENSE ==
- */
-
```
cbwebdevelopment: Love seeing these go away :)

src/components/trends/common/TrendsSVGContainer.js
```diff
@@ -1,3 +1,20 @@
+const BUMPERS = {
```
cbwebdevelopment: These constants should probably be moved below the import statements

src/components/trends/smbg/SMBGRangeAvgContainer.js
```diff
@@ -1,22 +1,7 @@
...
- *
- * You should have received a copy of the License along with this program; if
- * not, you can obtain one from Tidepool Project at tidepool.org.
- * == BSD2 LICENSE ==
- */
-
 import _ from 'lodash';
-import React, { PropTypes, PureComponent } from 'react';
+import PropTypes from 'prop-types';
+
```
cbwebdevelopment: Total nit-pick here.  It seems that for a number of these the codemod added a space between these 2 imports.  Might be nice to do a global find/replace on these to get rid of the space.

Up to you though.  Obviously is completely functional as-is.

test/components/common/stat/Stat.test.js
```diff
@@ -308,6 +308,7 @@ describe('Stat', () => {
       });
 
       it('should render the summary data when `isOpened` state is `false`', () => {
+        wrapper.setState({ isOpened: true });
```
cbwebdevelopment: As discussed, I was expecting to see these doubles throughout the tests.  I'm guessing you tried it, but in case not have you tried running `wrapper.update()` instead of the 2nd `setState` call?  (I'd test locally myself, but there's currently an npm registry issue preventing me from building)

krystophv: Yeah, unfortunately `wrapper.update()` didn't do anything. Having a breakpoint just after the first `setState` shows that it, indeed, doesn't actually set the state while the second round does. I'm afraid that, barring digging into `enzyme` code directly, it's going to be difficult to pin the cause of that down.

src/components/trends/common/TrendsSVGContainer.js
krystophv: Oh yeah they should.

src/components/trends/smbg/SMBGRangeAvgContainer.js
krystophv: Can do

PR #203, opened 2020-05-18
[WEB-559] [WEB-818] react upgrade/component library

krystophv:
[WEB-818] - merge react upgrade into develop to provide access to [WEB-559] component library on main development branch

[WEB-818]: https://tidepool.atlassian.net/browse/WEB-818
[WEB-559]: https://tidepool.atlassian.net/browse/WEB-559


test/components/common/stat/Stat.test.js
```diff
@@ -1762,7 +1762,7 @@ describe('Stat', () => {
         index: 1,
       });
 
-      sinon.assert.callCount(setStateSpy, 1);
+      // sinon.assert.callCount(setStateSpy, 1);
```
cbwebdevelopment, line 1765:
```suggestion
```

test/components/common/stat/Stat.test.js
```diff
@@ -1782,7 +1782,7 @@ describe('Stat', () => {
 
       instance.setChartTitle();
 
-      sinon.assert.callCount(setStateSpy, 1);
+      // sinon.assert.callCount(setStateSpy, 1);
```
cbwebdevelopment, line 1785:
```suggestion
```

test/components/common/stat/Stat.test.js
```diff
@@ -1799,7 +1799,7 @@ describe('Stat', () => {
...
```
cbwebdevelopment, line 1802:
```suggestion
```

src/components/settings/common/PumpSettingsContainer.js
```diff
@@ -1,3 +1,5 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
krystophv: Yeah, the codemods did a lot of that sort of thing. Tried to find all the instances with some global finds, but inevitably some slip through. Thanks for the catch

krystophv:
@cbwebdevelopment I'm good with that workaround. Since it seems to only affect the unit tests and not the user experience, I think we can proceed with that for now, but it does serve as an indicator that there's some fiddly bits with relying on the `UNSAFE` deprecations and that we'll have some work ahead of us when it comes time to convert over (it seems they're committed to supporting them at least through 17.x, so we can effectively kick the can down the road for a while)

PR #226, opened 2020-08-25
Release 1.14.0 to master

krystophv:
Associated PRs:
blip: https://github.com/tidepool-org/blip/pull/794
platform-client: https://github.com/tidepool-org/platform-client/pull/115


cbwebdevelopment (changes requested):
It's good to go, except I think we should bump the version to a production tag.

PR #232, opened 2020-09-29
[WEB-981] Customizable print date ranges

cbwebdevelopment:
See [WEB-981] for details.

This involved updating the PDF logic to optimize chart placements in order to, for instance, prevent a basics calendar from rendering over page breaks.

Also, the storybooks are updated to leverage the PDF data queries from blip.

Related PR: tidepool-org/blip#829

[WEB-981]: https://tidepool.atlassian.net/browse/WEB-981

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "1.14.0",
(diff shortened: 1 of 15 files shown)


thacherhussain (approved):
LGTM 🦄

PR #238, opened 2020-11-04
Add validation for automated boluses (WEB-775)

gniezen:
See [WEB-775] for more details.

Goes with:
- https://github.com/tidepool-org/tideline/pull/412

[WEB-775]: https://tidepool.atlassian.net/browse/WEB-775

### DIFF
--- a/src/utils/validation/schema.js
+++ b/src/utils/validation/schema.js
@@ -120,23 +120,7 @@ const normalBolus = {
   duration: forbidden,
   extended: forbidden,
   expectedExtended: forbidden,
-  subType: { type: 'string', enum: ['normal'] },
-  wizard: { type: 'string', pattern: patterns.id, ...optional },
-};
-
-const automatedBolus = {
-  ...common,
-  normal: minZero,
-  expectedNormal: {
-    type: 'withDependantFields',
-    schema: { ...minZero, ...optional },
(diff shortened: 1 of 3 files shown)


src/utils/validation/schema.js
```diff
@@ -124,6 +124,22 @@ const normalBolus = {
   wizard: { type: 'string', pattern: patterns.id, ...optional },
 };
 
+const automatedBolus = {
```
cbwebdevelopment, line 127: Looks like the only difference between this and the `normalBolus` schema is the `subType`.  Would it make more sense to simply add `automated` to the `subType` enum for `normalBolus`?

If we can do that, it will also reduce (or rather, not increase) the number of schemas we need to compare against for bolus datums.

cbwebdevelopment (changes requested):
This would work, but I'm wondering if we should avoid adding a new schema type to check against in this instance.

As well, whatever we do decide, we should also add the corresponding test updates in `schema.test.js`

src/utils/validation/schema.js
gniezen: Agreed, that makes it much simpler.

cbwebdevelopment (approved):
LGTM :shipit:

PR #253, opened 2021-02-23
[WEB-717] De-convert carb exchanges for Medtronic 5- and 7-series pumps

cbwebdevelopment:
It was discovered during testing of [WEB-717] that uploads for the Medtronic 5 and 7 series carb exchange data was converting the exchanges to carbs (and then rounding to nearest integer) at a 1:15 ratio.  

This conversion was intended to be removed when `carbUnits` was added to the data model, but it was missed.  We've decided to de-convert it on the frontend for the time being.  As well, the `insulinCarbRatio` on wizard datums needs de-conversion as well.  In that case, due to rounding and the very granular precision that might be used to input the ratios on the pump, there is potential irrecoverable precision loss.  For this, we add an annotation to the bolus tooltip for any carb ratios that could be affected.  The carb ratio on the settings view is fully precise.

There is a backend ticket created to fix the data and begin de-converting on the backend instead, at which point we will stop doing it on the frontend.

This also contains a couple of extra changes for the exchange data to show correctly in the copy-to-text button output for the basics and settings views.

[WEB-717]: https://tidepool.atlassian.net/browse/WEB-717

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "1.19.0-rc.5",
+  "version": "1.19.0-rc.6",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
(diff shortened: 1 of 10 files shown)


krystophv (approved):
LGTM 🚀

PR #255, opened 2021-03-09
Adding Microtech Equil pump to Device Settings (WEB-1145)

gniezen:
Addresses [WEB-1145].

[WEB-1145]: https://tidepool.atlassian.net/browse/WEB-1145


cbwebdevelopment (changes requested):
Nice work @gniezen.  I think you got it all covered here.  

My only suggestions are to make it a minor rather than major version bump in the `package.json`, and add a few unit tests to the corresponding test files for these changes.

I do have a couple questions as well:
1. Does the default `pumpVocabulary` in `constants.js` match the nomenclature used in the Equil pumps?  If not, we should set up a custom vocabulary like we have for the others.  If so, we're all good there.
2. Have you tested that the print view is working fine?  I normally would create a Storybook entry for new device settings for both the web and print views as my primary course of development.  It's fine not to have them, but wanted to be sure that the print view is working as expected.  I _think_ it would, but without having a way to see it, I can't be sure.

gniezen:
> My only suggestions are to make it a minor rather than major version bump in the `package.json`, and add a few unit tests to the corresponding test files for these changes.

Sorry, the major version  bump was a typo :man_facepalming: . Unit tests sound like a good idea, I'll work on that!

> 1. Does the default `pumpVocabulary` in `constants.js` match the nomenclature used in the Equil pumps?  If not, we should set up a custom vocabulary like we have for the others.  If so, we're all good there.

It looks fine, except for `Change Cartridge` which should be `Pump rewind`. Where is `SITE_CHANGE_RESERVOIR` shown in Web? Also, as this is a patch pump there's no "fill tubing".

> 2. Have you tested that the print view is working fine?  I normally would create a Storybook entry for new device settings for both the web and print views as my primary course of development.  It's fine not to have them, but wanted to be sure that the print view is working as expected.  I _think_ it would, but without having a way to see it, I can't be sure.

Print View looks good: 
image

cbwebdevelopment:
Cool 👍 
Re: `SITE_CHANGE_RESERVOIR`, it's used to determine the site change source data to show on the basics view (print and web - and possibly in the 'copy as text' output as well).  Perhaps this is the one that should be set for equil, since we use it for the Omnipod, as can be seen in the `getSiteChangeSource` function in `/src/utils/basics/data.js`

cbwebdevelopment (approved):
Great work on this, @gniezen!  LGTM :shipit:

PR #261, opened 2021-04-09
[WEB-775] Filter out non-control-iq upload data if control-iq data is present

cbwebdevelopment:
Part of [WEB-775].

This PR is to exclude pre-control-iq tandem uploads by default if we have data from the same device from a version of uploader supports control-iq data. 

Otherwise, we have duplicate data - specifically because the 'shape' of the bolus and basal datums changed due to the 'automated' subTypes, which in turn cause the generated datum id's to differ, and jellyfish deduplication does not happen.

To distinguish between data from uploads from a control-iq-capable pump prior to the adding of the automated subTypes, on the latest uploader version, we add begin the deviceId with `tandemCIQ` instead of just `tandem`, and, if we have, for instance, a `tandemCIQ123456789` device, we filter out, if present, by default,  the `tandem123456789` device.

[WEB-775]: https://tidepool.atlassian.net/browse/WEB-775

### DIFF
--- a/data/types.js
+++ b/data/types.js
@@ -30,7 +30,7 @@ export const generateGUID = () => 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace
 
 class Common {
   constructor(opts = {}) {
-    this.deviceId = 'Test Page Data - 123';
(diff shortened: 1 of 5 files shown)


krystophv (approved):
LGTM :shipit:

PR #284, opened 2021-12-01
Allow DataUtil storybook to use raw data from the API or blip console downloads

clintonium-119:
This is to allow us to use a variety of download sources in the DataUtil storybook, rather than just the private `accountTool.py` exports.

Main driving factor is a clinic IT team that would like to play with it.

Related PR: https://github.com/tidepool-org/blip/pull/971


krystophv (approved):
LGTM 👍🏼

PR #364, opened 2023-08-09
[WEB-2503,WEB-2504,WEB-2506] Fix bugs in matched device generation

clintonium-119:
[WEB-2503], [WEB-2504], [WEB-2506] 
Related PR: https://github.com/tidepool-org/blip/pull/1272

The previous method of determining matched devices as datums were normalized was insufficient.

It left many edge cases, especially for stat and aggregations, where it would work in some cases where data happened to be normalized as part of the stat generation process, but it wasn't guaranteed.

This now uses a crossfilter change handler that will generate the matched devices whenever the `byType` filter is changed to any datum types we visualize.  I'm using a `matchDevices` property on the DataUtil to control when this behavior occurs since we only want to do this while generating data for the `current` range period, and not the `prev` and `next` ones that get generated for some of the data views.

We also needed to control when we clear the matched devices, since we don't always query for the chart/stat/aggregation data in the same calls.

One final change that I implemented during the course of development was to remove some redundant `byDeviceId` excluded devices filters in the `AggregationUtil`, since they will always be already filtered out in `DataUtil.query` prior to the individual aggregations being generated.

[WEB-2503]: https://tidepool.atlassian.net/browse/WEB-2503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEB-2506]: https://tidepool.atlassian.net/browse/WEB-2506?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WEB-2504]: https://tidepool.atlassian.net/browse/WEB-2504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@tidepool/viz",
-  "version": "1.35.0-rc.1",
+  "version": "1.35.0-rc.2",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
--- a/src/utils/AggregationUtil.js
+++ b/src/utils/AggregationUtil.js
@@ -44,7 +44,6 @@ export class AggregationUtil {
 
   aggregateBasals = group => {
     this.dataUtil.filter.byType('basal');
-    this.dataUtil.filter.byDeviceIds(this.excludedDevices);
 
     let reducer = reductio();
     reducer.dataList(true);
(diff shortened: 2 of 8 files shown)


krystophv (approved):
LGTM :shipit:

PR #376, opened 2024-01-17
[WEB-2750] AGP cgm sufficiency check fix

clintonium-119:
See [WEB-2750]

This is a regression from when some stat properties used in the CGM sufficiency calculations were moved from the `sensorUsage` stat to `bgExtents`

Related PR: https://github.com/tidepool-org/blip/pull/1329


krystophv (approved):
LGTM 💯

PR #415, opened 2024-10-30
[WEB-3251] Set latestTimeZone metadata in data worker upon ingestion

clintonium-119:
[WEB-3251] Will go out as part of `1.81.1` hotfix

Moved the general logic for determining the ideal timezone based on data from blip to the data worker here, as well as including the latest `dosingDecision` datums, and `timeChange` device events to address Loop and Apple Health datasets falling back to browser timezone.

The data util seems to me like a better place to do this, and we would have had to send additional data out to blip anyhow in order to utlilze the latest time change data as a potential fallback for Apple Health-only datasets. 

Also added some missing unit tests for `DataUtil.setLatestDiabetesDatumEnd` that missed adding during `1.81.0` development.

[WEB-3251]: https://tidepool.atlassian.net/browse/WEB-3251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "20.8.0"
   },
   "packageManager": "yarn@3.6.4",
-  "version": "1.42.1-web-3250-tandem-insulin-duration-fix.1",
+  "version": "1.42.1-web-3251-use-dosing-decision-timezone-offset.1",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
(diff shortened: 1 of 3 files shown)


krystophv (approved):
LGTM 🚢

PR #433, opened 2025-01-14
[WEB-3179] dep updates

krystophv:
The last of the first round of deep dependency updates. Much more detail in the ticket for [WEB-3179], but had to navigate through multiple layers of breaking major version updates in `victory-charts` in order to get to a `d3` subdependency version that no longer had a security issue associated with it. While I was heading to the fridge anyway, I moved `victory-charts` to the latest version instead of one major version behind latest.

paired with blip pr: https://github.com/tidepool-org/blip/pull/1499

[WEB-3179]: https://tidepool.atlassian.net/browse/WEB-3179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "20.8.0"
   },
   "packageManager": "yarn@3.6.4",
-  "version": "1.44.0",
+  "version": "1.44.0-web-3179-dep-updates.1",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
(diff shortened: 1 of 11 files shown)


clintonium-119 (approved):
LGTM :100: 
I had taken a quick stab at updating victory and victory-core in last year's dep updates push, but gave up on it after seeing what a mess it would be to straighten out.

Nice work getting it all sorted.

PR #441, opened 2025-03-10
[WEB-3428,WEB-3472] Twiist site change and loop-specific rendering

clintonium-119:
[WEB-3428] [WEB-3472]

Related PRs:
https://github.com/tidepool-org/blip/pull/1533
https://github.com/tidepool-org/tideline/pull/459

This basically adds the ability to detect Twiist Loop based on the client name and major version (>2)

It also adds the ability to add `loop` tags to `food` and `bolus` datums.  Historically, we've had run the datums through the various `isLoop` utilities directly at runtime, but this is needed for twiist data, since only the upload contains a cilent name we can compare against.  It also means that we'll be able to tell if a datum is loop or not outside of `viz`, which may prove useful.

Also adds the twiist site changes, and sets defaults for the devices that present different site change source options.

[WEB-3428]: https://tidepool.atlassian.net/browse/WEB-3428?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEB-3472]: https://tidepool.atlassian.net/browse/WEB-3472?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


test/utils/device.test.js
```diff
@@ -144,6 +144,25 @@ describe('device utility functions', () => {
...
+
+    it('should return `false` for a non-matching pattern within `client.name`', () => {
+      const datum = { client: { name: 'com.tidepool.Loop', version: '2.0.0' } };
+      expect(device.isTwiistLoop(datum)).to.be.false;
+    });
+
+    it('should return `false` for a client major version below 2', () => {
+      const datum = { client: { name: 'com.tidepool.Loop', version: '1.9.9' } };
+      expect(device.isTwiistLoop(datum)).to.be.false;
+    });
```
krystophv, line 163: This test has the same `client.name` as the test for non-matching `client.name`, so may not actually be testing the major version restriction.

clintonium-119: Good catch :)  Fixing.

krystophv (approved):
LGTM 🚀

PR #443, opened 2025-03-13
WEB-2965 - Exclude Days with no insulin data in Basics

henry-tp:
[WEB-2965]

[WEB-2965]: https://tidepool.atlassian.net/browse/WEB-2965?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/StatUtil.js
```diff
@@ -82,6 +84,20 @@ export class StatUtil {
...
 
+    const uniqueDatumDates = new Set();
+
+    bolusData.forEach(datum => {
+      const date = formatLocalizedFromUTC(datum.time, this.timePrefs, 'YYYY-MM-DD');
+      uniqueDatumDates.add(date);
+    });
+
+    basalData.forEach(datum => {
+      const date = formatLocalizedFromUTC(datum.time, this.timePrefs, 'YYYY-MM-DD');
```
henry-tp, line 95: NTS: figure out the `overlap` situation

src/utils/StatUtil.js
```diff
@@ -82,6 +84,20 @@ export class StatUtil {
...
+
+    bolusData.forEach(datum => {
+      const date = formatLocalizedFromUTC(datum.time, this.timePrefs, 'YYYY-MM-DD');
+      uniqueDatumDates.add(date);
+    });
+
+    rawBasalData.forEach(datum => { // use rawBasalData to avoid counting dates of overlap
+      const date = formatLocalizedFromUTC(datum.time, this.timePrefs, 'YYYY-MM-DD');
+      uniqueDatumDates.add(date);
+    });
```
henry-tp, line 97: My rationale for excluding the overlap is, that, given the following scenario where each purple box is a basal datum:

image

You don't want to count Monday or Friday even though they are part of the overlap. By using the startTime of each basal and not counting the overlapping basals, we should be able to restrict this to the 3 days of significance here

src/utils/StatUtil.js
```diff
@@ -82,16 +83,24 @@ export class StatUtil {
     const rawBasalData = this.dataUtil.sort.byTime(this.dataUtil.filter.byType('basal').top(Infinity));
     const basalData = this.dataUtil.addBasalOverlappingStart(_.cloneDeep(rawBasalData));
 
+    // Create a list of all dates for which we have at least one datum
+    const uniqueDatumDates = new Set([
+      ...bolusData.map(datum => formatLocalizedFromUTC(datum.time, this.timePrefs, 'YYYY-MM-DD')),
+      ...rawBasalData.map(datum => formatLocalizedFromUTC(datum.time, this.timePrefs, 'YYYY-MM-DD')),
```
henry-tp, line 89: I understand each basal datum to have a time range with `normalTime` and `endTime` being the start and end times respectively. From what I can see, each datum

src/utils/StatUtil.js
```diff
@@ -82,16 +83,24 @@ export class StatUtil {
...
+    ]);
```
henry-tp, line 90: - I think this code is acceptably readable, but if you think it's a bit too obfuscated let me know and I can make it a bit more imperative

- I understand each basal datum to have a time range with `normalTime` and `endTime` being the start and end times respectively. From what I can see, each datum

henry-tp: As for using `rawBasalData` instead of `basalData`, my thought process is - given the following scenario where each purple box is a basal datum:

image

My understanding is that each basal datum is ~5 minutes, so we don't want to count Monday or Friday. By using the startTime of each basal and not counting the overlapping basals, we should be able to restrict this to the 3 days of significance here. I think.

PR #447, opened 2025-03-24
WEB-3418 - Use Redirect instead of Pop-up when connecting Provider on Mobile

henry-tp:
[WEB-3418]

Associated with https://github.com/tidepool-org/blip/pull/1540

[WEB-3418]: https://tidepool.atlassian.net/browse/WEB-3418?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/components/common/stat/Stat.js
```diff
@@ -19,9 +19,9 @@ import BgBar from './BgBar';
 import BgBarLabel from './BgBarLabel';
 import StatTooltip from '../tooltips/StatTooltip';
 import StatLegend from './StatLegend';
-import CollapseIconOpen from './assets/expand-more-24-px.svg';
-import CollapseIconClose from './assets/chevron-right-24-px.svg';
-import InfoIcon from './assets/info-outline-24-px.svg';
+import CollapseIconOpen from './assets/expand-more-24-px.png';
+import CollapseIconClose from './assets/chevron-right-24-px.png';
+import InfoIcon from './assets/info-outline-24-px.png';
```
henry-tp, line 24: See this comment: https://github.com/tidepool-org/blip/pull/1540/files#r2010755798

clintonium-119: This is fine, though I think we should remove the svg icons, and update any other uses to use the png just to keep things simple/consistent.

I think it's only the info-outline icon that's reused, but you'll need to check.

clintonium-119 (changes requested):
Just one small change request in comments - removing the old svg icons

clintonium-119 (approved):
LGTM :ship:

PR #454, opened 2025-04-03
WEB-3501 - Adjust Thresholds for Displayed Ranges

henry-tp:
[WEB-3501]

Blip: https://github.com/tidepool-org/blip/pull/1551

Tideline: https://github.com/tidepool-org/tideline/pull/471

[WEB-3501]: https://tidepool.atlassian.net/browse/WEB-3501?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/utils/stat.js
```diff
@@ -207,14 +207,7 @@ export const formatDatum = (datum = {}, format, opts = {}) => {
...
         value = _.max([value, 0]);
-        const percentage = (value / total) * 100;
-        let precision = 0;
-        // We want to show extra precision on very small percentages so that we avoid showing 0%
-        // when there is some data there.
-        if (percentage > 0 && percentage < 0.5) {
-          precision = percentage < 0.05 ? 2 : 1;
-        }
-        value = formatDecimalNumber(percentage, precision);
+        value = formatStatsPercentage(value / total);
```
henry-tp, line 210: As far as I can tell, the `commonStats` that this impacts are:

`readingsInRange`
`sensorUsage`
`timeInAuto`
`timeInOverride`
`timeInRange`
`totalInsulin`

From my assessment, this change _should be_ safe.

The only one I'm really confused about is `readingsInRange`, which is `Number of Readings` in this AGP. It seems like it should be defined as a discrete value instead of a percentage? It still seems safe, though.

image

src/utils/bloodglucose.js
```diff
@@ -42,25 +43,42 @@ export function classifyBgValue(bgBounds, bgValue, classificationType = 'threeWa
...
+
+      return 'target';
+
+    case 'threeWay':
+      if (bgValue < targetLowerBound) {
+        return 'low';
+      } else if (bgValue > targetUpperBound) {
+        return 'high';
+      }
+      return 'target';
```
henry-tp, line 80: threeWay is unchanged

src/components/daily/cbgtooltip/CBGTooltip.js
```diff
@@ -68,6 +69,7 @@ class CBGTooltip extends PureComponent {
   render() {
     const bgClass = classifyBgValue(
       reshapeBgClassesToBgBounds(this.props.bgPrefs),
+      this.props.bgPrefs.bgUnits,
```
henry-tp, line 72: Just for awareness, the tooltips may in edge cases be misaligned with Tideline's internal fiveway classification now. 

See these two images from PRESENT STATE of the platform:

| 1 | 2 |
|-----|-----|
| image | image |

The plotted points come from Tideline but the tooltips from Viz. So even though the data points are both 10.0, there is internal rounding that is not reflected. But at least they are consistent in Tideline and Viz.

Now, after this PR, the same data points:

| 1 | 2 |
|-----|-----|
| image | image |

As you can see we have achieved internal consistency within Viz but not Tideline. I guess Tideline doesn't import Viz so it's harder to rectify this without copying and pasting all of the `classifyBgValue` logic to Tideline as well.

src/utils/bloodglucose.js
```diff
@@ -42,25 +43,42 @@ export function classifyBgValue(bgBounds, bgValue, classificationType = 'threeWa
...
+        return 'veryLow';
+      } else if (bgValue > veryHighThreshold) {
+        return 'veryHigh';
+      }
+
+      // Low, Target, and High ranges are non-contiguous in the ADA Standardized CGM metrics.
+      // We ensure that values falling between these ranges are rounded into an appropriate
+      // range before trying to classify them. See the unit tests for examples of scenarios.
+      const precision = bgUnits === MMOLL_UNITS ? 1 : 0;
+      const roundedValue = bankersRound(bgValue, precision);
```
henry-tp, line 64: Here, we go from ONLY using `<` or `>` comparisons to using `bankersRound` potentially on every single data point, so I was a little worried about performance. According to my testing, running 300k data points through bankersRound only adds ~4ms.

src/utils/bloodglucose.js
```diff
@@ -126,14 +144,20 @@ export function reshapeBgClassesToBgBounds(bgPrefs) {
  */
 export function generateBgRangeLabels(bgPrefs, opts = {}) {
   const { bgBounds, bgUnits } = bgPrefs;
+  const minimumIncrement = BG_DISPLAY_MINIMUM_INCREMENTS[bgUnits];
+
   const thresholds = _.mapValues(bgBounds, threshold => formatBgValue(threshold, bgPrefs));
+  thresholds.highLowerBound = formatBgValue(bgBounds.targetUpperBound + minimumIncrement, bgPrefs);
+  thresholds.lowUpperBound = formatBgValue(bgBounds.targetLowerBound - minimumIncrement, bgPrefs);
```
henry-tp, line 151: We would have to keep this approach in mind when we get to custom ranges. For the ADA ranges, we get `10.0` as the end of target, and I can say that "high" begins at `10.0` + `0.1`. But if we're not following ADA ranges anymore, is it guaranteed to be `targetUpperBound + 0.1`? If the backend says `targetUpperBound` is `11.0`, is the beginning of "high" certainly `11.1`?

src/utils/bloodglucose.js
```diff
@@ -42,19 +43,35 @@ export function classifyBgValue(bgBounds, bgValue, classificationType = 'threeWa
...
+    // We ensure that values falling between these ranges are rounded into an appropriate
+    // range before trying to classify them. See the unit tests for examples of scenarios.
+    const precision = bgUnits === MMOLL_UNITS ? 1 : 0;
+    const roundedValue = bankersRound(bgValue, precision);
+
+    if (roundedValue < targetLowerBound) {
+      return 'low';
+    } else if (roundedValue > targetUpperBound) {
+      return 'high';
+    }
```
henry-tp, line 69: Here, we go from ONLY using `<` or `>` comparisons to using `bankersRound` potentially on every single data point, so I was a little worried about performance. According to my testing, running 300k data points through bankersRound only adds ~4ms, so no concerns there.

src/utils/bloodglucose.js
```diff
@@ -42,19 +43,35 @@ export function classifyBgValue(bgBounds, bgValue, classificationType = 'threeWa
...
     if (bgValue < veryLowThreshold) {
       return 'veryLow';
-    } else if (bgValue >= veryLowThreshold && bgValue < targetLowerBound) {
-      return 'low';
-    } else if (bgValue > targetUpperBound && bgValue <= veryHighThreshold) {
-      return 'high';
     } else if (bgValue > veryHighThreshold) {
       return 'veryHigh';
     }
+
```
henry-tp, line 58: Again, to re-emphasize, there is a difference now between this implementation and what is described here: https://github.com/tidepool-org/TidepoolApi/blob/master/docs/summary-data.md

Is it the right thing to do? I dont' know 😭 😭 😭

clintonium-119 (changes requested):
Great job.  I did have a small suggestions/questions, but this looks great.
It may not even require a change on your part, in which case I'll just approve this.

src/utils/bloodglucose.js
clintonium-119: Well, isn't that... something :laughing:

clintonium-119 (approved):
LGTM :100:

PR #466, opened 2025-04-17
Fix Override/Underride Calculations

henry-tp:
[WEB-3399]

Based on feedback from this [discussion thread](https://tidepoolteam.slack.com/archives/CFSURMMS9/p1744769521450649)

I thought i would just simplify the logic down since we don't really need to look at the precision anymore

[WEB-3399]: https://tidepool.atlassian.net/browse/WEB-3399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/bolus.js
+++ b/src/utils/bolus.js
@@ -333,18 +333,6 @@ export function isInterruptedBolus(insulinEvent) {
 }
 
 /**
- * isDifferentBeyondPrecision
- * @param {number} a - The first number to compare.
- * @param {number} b - The second number to compare.
- * @param {number} precision - The number of decimal places to consider.
- *
- * @returns {boolean} Whether the numbers are different when rounded to the specified precision.
- */
-export function isDifferentBeyondPrecision(a, b, precision) {
-  return _.round(a, precision) !== _.round(b, precision);
-}
-
-/**
  * isOverride
  * @param {Object} insulinEvent - a Tidepool bolus or wizard object
  *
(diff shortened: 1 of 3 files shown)


src/utils/bolus.js
```diff
@@ -351,10 +351,11 @@ export function isDifferentBeyondPrecision(a, b, precision) {
  * @return {Boolean} Whether the programmed amount is both significantly different (beyond 2 decimal places) from and larger than the recommended amount.
  */
 export function isOverride(insulinEvent) {
+  const MINIMUM_THRESHOLD = 0.01;
   const amountRecommended = getRecommended(insulinEvent.wizard || insulinEvent.dosingDecision || insulinEvent);
   const amountProgrammed = getProgrammed(insulinEvent);
 
-  return isDifferentBeyondPrecision(amountRecommended, amountProgrammed, 2) && amountProgrammed > amountRecommended;
```
clintonium-119, line 357: I'd recommend that we also delete the `isDifferentBeyondPrecision` function and it's tests, and add a few tests to isOverride and isUnderride to assert that differences below the min threshold are handled correctly.

clintonium-119 (changes requested):

src/utils/bolus.js
henry-tp: Thanks @clintonium-119 . I have removed the old code and added some test cases. Please have a look again, thanks!

clintonium-119 (approved):
LGTM :100:

PR #475, opened 2025-05-16
[WEB-3677] Allow filtering by cgm sample interval

clintonium-119:
[WEB-3677]

### DIFF
--- a/src/utils/DataUtil.js
+++ b/src/utils/DataUtil.js
@@ -27,7 +27,7 @@ import {
   isUnderride,
 } from './bolus';
 
-import { convertToMGDL, defaultCGMSampleInterval } from './bloodglucose';
+import { convertToMGDL } from './bloodglucose';
 
 import {
   BGM_DATA_KEY,
@@ -79,6 +79,8 @@ export class DataUtil {
     this.startTimer('init total');
     this.data = crossfilter([]);
     this.queryDataCount = 0;
+    this.defaultCGMSampleInterval = 5 * MS_IN_MIN;
+    this.defaultCGMSampleIntervalRange = [this.defaultCGMSampleInterval, Infinity];
 
     this.buildDimensions();
     this.buildFilters();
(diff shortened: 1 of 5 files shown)


src/utils/DataUtil.js
```diff
@@ -203,6 +207,20 @@ export class DataUtil {
       d.time = d.timestamp;
     }
 
+    if (d.type === 'cbg' && !d.sampleInterval) {
```
clintonium-119, line 210: This is basically the old `cgmSampleFrequency` utility transplanted here, with the `sampleInterval` being added to each cgm datum upon ingestion for legacy data that does not have it.

This is needed since we need to use a new crossfilter dimension to allow us to efficiently filter by `sampleInterval`, so each datum needs a value set.

src/utils/StatUtil.js
```diff
@@ -33,7 +33,13 @@ export class StatUtil {
     this.log('bgPrefs', { bgBounds: this.bgBounds, bgUnits: this.bgUnits });
   };
 
+  filterCBGDataByDefaultSampleInterval = () => {
```
clintonium-119, line 36: All generated stats will use the default sample interval range for calculations, even if the user is viewing 1min data

src/utils/bloodglucose.js
```diff
@@ -234,48 +234,15 @@ export function getOutOfRangeThreshold(bgDatum) {
  * @param {Array} data - cgm data
  * @return {Integer} count - the weighted count
  */
-export function weightedCGMCount(data) {
+export function weightedCGMCount(data) { // TODO: do I need to account for 1min samples here?
```
clintonium-119, line 237: Now that each datum has a sampleInterval set, we can use that for the weighted count (currently only used for determining the default bgSource for the trends view)

test/utils/device.test.js
```diff
@@ -295,6 +295,20 @@ describe('device utility functions', () => {
     });
   });
 
+  describe('isOneMinCGMSampleIntervalDevice', () => {
+    it('should return `true` for an upload record for a device upload with one minute cgm interval capabilities', () => {
+      expect(device.isOneMinCGMSampleIntervalDevice({ type: 'upload', client: { name: 'com.sequelmedtech.tidepool-service', version: '2.0.0' } })).to.be.true;
+      expect(device.isOneMinCGMSampleIntervalDevice({ origin: { name: 'com.dekaresearch.twiist' } })).to.be.true;
+    });
+
+    it('should return `false` for an upload record for a device upload with one minute cgm interval capabilities', () => {
```
krystophv, line 304:
```suggestion
    it('should return `false` for an upload record for a device upload without one minute cgm interval capabilities', () => {
```

PR #481, opened 2025-06-09
[WEB-3688] Generate suppressed basal data for twiist

clintonium-119:
[WEB-3688]

[WEB-3688]: https://tidepool.atlassian.net/browse/WEB-3688?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/data/types.js
+++ b/data/types.js
@@ -241,6 +241,7 @@ export class Settings extends Common {
     this.deviceTime = opts.deviceTime;
     this.insulinSensitivity = opts.insulinSensitivity;
     this.units = opts.units;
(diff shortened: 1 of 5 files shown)


krystophv (changes requested):
This was a _read_ - the comments helped a lot, but it's still doing a really complicated thing. Excellent work. There's a couple small things in unit tests I think we should fix, though.

PR #498, opened 2025-07-04
[WEB-3800][WEB-3801] tag wizards

krystophv:
Takes care of [WEB-3800] and [WEB-3801]
Since wizard rendering in tideline is now [looking for tags](https://github.com/tidepool-org/tideline/pull/477/files#diff-52851a90d24200b579e99591495b25d70e644f33b96a07424f544247ed28bdfc), we should tag em.

### DIFF
--- a/src/utils/DataUtil.js
+++ b/src/utils/DataUtil.js
@@ -624,6 +624,20 @@ export class DataUtil {
       };
     }
 
+    if (d.type === 'wizard') {
+      d.tags = {
+        automated: isAutomated(d),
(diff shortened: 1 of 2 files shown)


clintonium-119 (approved):
Ha :)  Yes, I should probably have made sure that we were tagging wizards!

LGTM :1st_place_medal: 

I did limit the tags we set to only include those we use in tideline, as some of the others weren't really relevant and/or possible (i.e. a loop wizard doesn't, and will likely never, exist)

PR #532, opened 2025-08-30
WEB-3925 - Align FE Summary Stat Salculations with BE

henry-tp:
[WEB-3925]

blip: https://github.com/tidepool-org/blip/pull/1740

## Manual Testing Discussion:

For testing purposes, I asked Brian to create a custom patient for me, and I recommend you take a look at them too.

Env: QA1
Clinic: `68bf26c66093f40937294344` (name: "source")
Patient: `fbcb8ae5-7b8c-46da-9377-dbb82347aa82` (name: "patient 9112025d")
- can be accessed by the clinic user `brian+upload786web1327clinic35a@tidepool.org` /1p

-----

The logic for data population for this patient is as follows:
- The patient has 2 CGM devices
- Device 1 reports only 52mg/dL with sampleInterval of 5 minutes with no missing data
- Device 2 reports only 182 mg/dL with sampleInterval of 5 minutes with no missing data
- Device 2 reports readings that are offset to those Device 1 by 60 seconds (e.g. Device 1 has readings at 12:30, 12:35, and 12:40, then Device 2 has readings at 12:31, 12:36, and 12:41)

My logic is this: If the data is set up correctly and the algorithm is working as intended, then every single one of Device 2's readings will be ignored/discarded in deduplication (because they are never past the blackout window). This makes it much easier to verify that every stat block is showing a correct value, because the values should behave precisely as though all of the data comes from Device 1. Any small deviation (e.g. average glucose any higher than 52mg/dL, stdev any higher than 0) would mean that the algorithm is not working as intended

Here is a side-by-side comparison of this branch (left) vs `develop` (right). To me, I think it is having the intended effect, and the values are reflecting exactly what I would expect

image

image

You can see that this aligns with BE summary method (100% veryLow instead of 50/50):

image

For any other patients of course, the statistics values should basically be the exact same between this branch and develop.

[WEB-3925]: https://tidepool.atlassian.net/browse/WEB-3925?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "20.8.0"
   },
   "packageManager": "yarn@3.6.4",
-  "version": "1.51.0-web-3925-wear-time.6",
+  "version": "1.51.0-web-3925-wear-time.7",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"


src/utils/StatUtil.js
```diff
@@ -295,48 +295,55 @@ export class StatUtil {
...
+    // TODO: Comments
+    const potentialTotal = this.activeDays * MS_IN_DAY;
+    const realTotal = potentialTotal - (moment(lastRecord.time).endOf('hour').valueOf() - lastRecord.time)
+
+    // AGP Calculations
+    const lastRecordTime = lastRecord.time;
+    const firstRecordTime = cbgData[0]?.time;
+    const sampleIntervalMinutes = Math.round(lastRecord?.sampleInterval / MS_IN_MIN);
+    const cgmMinutesPossible = Math.ceil(((lastRecordTime - firstRecordTime) / MS_IN_MIN) / sampleIntervalMinutes) + 1;
+    const sensorUsageAGP = (totalRecords / cgmMinutesPossible) * 100;
```
wzeller, line 337: This won't work for data having mixed sample intervals.  For example, if the first half of the CGM data was 15-min and the last half was 5-min, then this method will show a very incorrectly low wear time.

You want to figure out how many 5-min values were possible between the start and end time, and then how many minutes of cgm data we saw (duration/MS_in_MIN * 5). Dividing minutes of cgm data / possible 5-min chunks would give you the correct sensorUsageAGP (still different than above, because it only looks at the time period between first and last cgm value).

henry-tp: Thanks @wzeller . I was trying to re-implement the AGP spec, but the formula doesn't quite factor in `sampleIntervals` that vary or missing data points.

image

I implemented this as you wrote, but found that the number 5 cancels out in the numerator and denominator. I guess this is will be more accurate for 1-minute CGM as well.

src/utils/StatUtil.js
```diff
@@ -295,48 +295,51 @@ export class StatUtil {
...
+      totalRecords += 1;
+      lastRecord = currentRecord;
+
+      // Then, set a new blackout window based on the current record's time window.
+      blackoutWindow = currentRecord.time + currentRecord.sampleInterval - OVERLAP_TOLERANCE;
+    };
+
+    // To mimic backend logic, we slice off the time until the end of the hour for the last reading
+    const potentialTotal = this.activeDays * MS_IN_DAY;
+    const realTotal = potentialTotal - (moment(lastRecord.time).endOf('hour').valueOf() - (lastRecord.time + lastRecord.sampleInterval));
```
henry-tp, line 330: Mimics the logic according to [this call](https://github.com/tidepool-org/platform-private/blob/master/data/summary/types/cgm.go#L800) to [this method](https://github.com/tidepool-org/platform-private/blob/master/data/summary/types/util.go) when calculating backend summary.

Admittedly, this does not make the most sense to me (see [slack discussion ](https://tidepoolteam.slack.com/archives/C087MHT9R/p1756851776424819?thread_ts=1756850926.081549&cid=C087MHT9R)), but I am doing my best to mimic BE logic here

src/utils/StatUtil.js
```diff
@@ -295,48 +295,55 @@ export class StatUtil {
...
+      blackoutWindow = currentRecord.time + currentRecord.sampleInterval - OVERLAP_TOLERANCE;
+    };
+
+    // To mimic backend logic, we slice off the time until the end of the hour for the last reading
+    const total = (() => {
+      const potentialTotal = this.activeDays * MS_IN_DAY;
+      const realTotal = potentialTotal - (moment(lastRecord.time).endOf('hour').valueOf() - (lastRecord.time + lastRecord.sampleInterval));
+
+      return _.min([realTotal, potentialTotal]);
+    })();
```
henry-tp, line 334: Mimics the logic according to [this call](https://github.com/tidepool-org/platform-private/blob/master/data/summary/types/cgm.go#L800) to [this method](https://github.com/tidepool-org/platform-private/blob/master/data/summary/types/util.go) when calculating backend summary.

Admittedly, this does not make the most sense to me (see [slack discussion](https://tidepoolteam.slack.com/archives/C087MHT9R/p1756851776424819?thread_ts=1756850926.081549&cid=C087MHT9R)), but I am doing my best to mimic BE logic here

src/utils/StatUtil.js
henry-tp: Thanks @wzeller . I had attempted to re-implement the formula in the AGP spec, but I can see that that formula doesn't handle varying CGM `sampleIntervals`.

If we do it this way, the calculation would be

```js
const end = lastRecord?.time + lastRecord?.sampleInterval; // end of last datum
const start = cbgData[0]?.time                             // start of first datum

const sensorUsageAGP = (
  (duration / (MS_IN_MIN * 5)) /                           // total number of 5-minute chunks
  ((end - start) / (MS_IN_MIN * 5))                        // total number of 5-minute chunks possible
);
```

Simplifying mathematically, we'd get

```js
const end = lastRecord?.time + lastRecord?.sampleInterval; // end of last datum
const start = cbgData[0]?.time                             // start of first datum

const sensorUsageAGP = duration / (end - start);
```

So it seems like it ends up deceptively simple?

src/utils/StatUtil.js
```diff
@@ -296,47 +296,49 @@ export class StatUtil {
...
-      sensorUsage: duration,
-      sensorUsageAGP,
-      total,
-      sampleInterval,
-      count,
+      sensorUsage:    duration,
+      sensorUsageAGP: sensorUsageAGP || 0,
+      total:          total,
+      sampleInterval: lastRecord?.sampleInterval || this.dataUtil.defaultCgmSampleInterval,
+      count:          totalRecords,
```
henry-tp, line 341: This code heavily leans into use of `sampleInterval` whereas the previous code did not. According to @clintonium-119 in [slack](https://tidepoolteam.slack.com/archives/C07F46SL6QZ/p1756931324968609?thread_ts=1756401790.242519&cid=C07F46SL6QZ), all datums here should be guaranteed to have a `sampleInterval`, so this code should be safe on that front

test/utils/StatUtil.test.js
```diff
@@ -843,10 +846,7 @@ describe('StatUtil', () => {
       expect(result.total).to.equal(MS_IN_DAY);
       expect(result.sampleInterval).to.equal(expectedSampleFrequency);
       expect(result.count).to.equal(expectedCount);
-      expect(result.sensorUsageAGP).to.equal((
-        expectedCount /
-        ((expectedCGMMinutesWorn / (expectedSampleFrequency / MS_IN_MIN)) + 1)
-      ) * 100);
```
henry-tp, line 849: The previous value here was `45.4545...`. Based on the AGP specification, I believe it should be 100% here. Per the AGP Spec,

```
(𝑡𝑚𝑁 − 𝑡𝑚1 ) should be calculated as the number of minutes between the first and last measurement
```

In this case it is 55 minutes (3 x 15 min from Libre, 2 x 5 min from Dexcom), of which every minute is populated by a reading. So I believe it should be 100%.

test/utils/StatUtil.test.js
```diff
@@ -843,10 +846,7 @@ describe('StatUtil', () => {
...
+      expect(result.sensorUsageAGP).to.equal(100);
```
henry-tp, line 849: The previous resultant value here was ~45.4545%. Based on the AGP specification, I believe it should be 100% here. Per the AGP Spec,

```
(𝑡𝑚𝑁 − 𝑡𝑚1 ) should be calculated as the number of minutes between the first and last measurement
```

In this case it is 55 minutes (3 x 15 min from Libre, 2 x 5 min from Dexcom), of which every minute is populated by a reading. So I believe 100% is correct according to specifications

test/utils/StatUtil.test.js
```diff
@@ -857,10 +857,7 @@ describe('StatUtil', () => {
       expect(result.total).to.equal(MS_IN_DAY * 14);
       expect(result.sampleInterval).to.equal(expectedSampleFrequency);
       expect(result.count).to.equal(expectedCount);
-      expect(result.sensorUsageAGP).to.equal((
-        expectedCount /
-        ((expectedCGMMinutesWorn / (expectedSampleFrequency / MS_IN_MIN)) + 1)
-      ) * 100);
+      expect(result.sensorUsageAGP).to.equal((3600000 / (1517529900000 - 1517443200000)) * 100); // ~ 4.15%
```
henry-tp, line 860: This was previously ~2.076%.

src/utils/StatUtil.js
```diff
@@ -296,47 +296,49 @@ export class StatUtil {
...
-      (result, datum) => {
-        result += datum.sampleInterval;
-        return result;
-      },
-      0
-    );
+    // Iterate through the time-ordered list of glucose data. For each datum, we set a blackout
+    // period for the window of time it should occupy - e.g. if a datum has a time of 14:30 and
+    // a duration of 5 mins, then any other records occuring between 14:30 to 14:35 should NOT
+    // contribute to the CGM wear time.
```
henry-tp, line 305: The logic here is borrowed from `https://github.com/tidepool-org/platform/blob/20a0b9e5f8f7fb480216976dbb6cb149d0343248/summary/types/glucose.go#L349-L368`

However, it is heavily simplified due to not needing to group data into 1-hour buckets as is necessary in BE for calculating other CGM stats

src/utils/StatUtil.js
```diff
@@ -296,47 +296,49 @@ export class StatUtil {
...
+      // Then, set a new blackout window based on the current record's time window.
+      blackoutWindow = currentRecord.time + currentRecord.sampleInterval - OVERLAP_TOLERANCE;
+    };
+
+    const total = this.activeDays * MS_IN_DAY;
+
+    // AGP Calculations
+    const end = lastRecord?.time + lastRecord?.sampleInterval;
+    const start = cbgData[0]?.time;
+    const sensorUsageAGP = (duration / (end - start)) * 100;
```
henry-tp, line 334: To be clear, this is now a modified sensorUsageAGP calculation that doesn't not follow the AGP spec precisely. The original one according to AGP spec has a singular `r` value (representing the rate of sampling or the `sampleInterval`), but the requirement here was to be able to handle readings from combined mixed 15 minute and 5 minute datums.

src/utils/StatUtil.js
```diff
@@ -296,47 +296,49 @@ export class StatUtil {
...
+      // If current record occured within the current blackout window, discard it
+      if (currentRecord.time < blackoutWindow) continue; // eslint-disable-line no-continue
+
+      // If the record is past the blackout window, add its duration to the total wear time.
+      duration += currentRecord.sampleInterval;
+      totalRecords += 1;
+      lastRecord = currentRecord;
+
+      // Then, set a new blackout window based on the current record's time window.
+      blackoutWindow = currentRecord.time + currentRecord.sampleInterval - OVERLAP_TOLERANCE;
```
henry-tp, line 326: This code heavily leans into use of `sampleInterval` whereas the previous code did not. According to @clintonium-119 in [slack](https://tidepoolteam.slack.com/archives/C07F46SL6QZ/p1756931324968609?thread_ts=1756401790.242519&cid=C07F46SL6QZ), all datums here should be guaranteed to have a `sampleInterval`, so this code should be safe on that front

src/utils/StatUtil.js
henry-tp: This logic _could_ be abstracted out to be used as a filtering method for datums in general in this StatUtil class if we want to do so.

src/utils/StatUtil.js
```diff
@@ -426,7 +430,8 @@ export class StatUtil {
 
   getTimeInRangeData = () => {
     this.filterCBGDataByDefaultSampleInterval();
-    const cbgData = _.cloneDeep(this.dataUtil.filter.byType('cbg').top(Infinity));
+    const rawCbgData = this.dataUtil.filter.byType('cbg').top(Infinity);
+    const cbgData = this.dataUtil.deduplicateCBGData(rawCbgData);
```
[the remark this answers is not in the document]
clintonium-119: See comment above.  I had run a bunch of scenarios through the DataUtil storybook, using console.timer for to capture each call execution time, and there are some scenarios where it really adds up.

src/utils/StatUtil.js
```diff
@@ -297,48 +307,40 @@ export class StatUtil {
...
+      cgmMinutesWorn = cbgData.length === 1 ? sampleInterval : 0;
+    } else {
+      cgmMinutesWorn = Math.ceil(moment.utc(newestDatum?.time).diff(moment.utc(oldestDatum?.time), 'minutes', true));
+    }
 
-    // AGP Calculations
-    const end = lastRecord?.time + lastRecord?.sampleInterval;
-    const start = cbgData[0]?.time;
-    const sensorUsageAGP = (duration / (end - start)) * 100;
+    const sensorUsageAGP = (
```
clintonium-119, line 333: Maybe I'm not recalling correctly, but I thought we were going to leave the AGP calculations untouched.

clintonium-119 (changes requested):
It's a good start, but there are performance concerns that should be ironed out, and a few other concerns/comments to address.

src/utils/DataUtil.js
```diff
@@ -1001,6 +1001,34 @@ export class DataUtil {
...
+      if (currentRecord.time < blackoutUntil) continue; // eslint-disable-line
+
+      // Otherwise, if the record is past the blackout window, we include the record.
+      // Then, set a new blackout window based on the current record's time window
+      output.push(currentRecord);
+      blackoutUntil = currentRecord.time + currentRecord.sampleInterval - OVERLAP_TOLERANCE;
+    }
+
+    return output;
+  };
```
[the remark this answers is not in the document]
henry-tp: Great catch on this Clint. I've memoized the call per your recommendation. Cool part is that if you flip _back_ in the calendar to a spot you've already viewed it doesn't need to call the function again too. Super cool.

982c6c90

src/utils/StatUtil.js
henry-tp: Oh interesting - I clicked on your comment, and I think what you were looking at was the modifications I made to _revert_ the AGP changes back to the original code (after our chat with Bill). If you look at the latest changes, the AGP calculations are in their original state

src/utils/AggregationUtil.js
```diff
@@ -549,7 +549,10 @@ export class AggregationUtil {
 
         // Filter cgm data by the currently-set sample interval range.
         this.dataUtil.filter.bySampleIntervalRange(...(this.dataUtil.cgmSampleIntervalRange || this.dataUtil.defaultCgmSampleIntervalRange));
-        groupedData.cbg = this.dataUtil.filter.byType('cbg').top(Infinity);
+
```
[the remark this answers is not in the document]
henry-tp: for discussion in meeting

src/utils/print/plotly.js
```diff
@@ -49,7 +49,10 @@ export async function generateAGPFigureDefinitions(data) {
 
   // Generate dailyGlucoseProfiles figures
   if (sections.dailyGlucoseProfiles.sufficientData) {
-    const bgDataByDate = _.mapValues(data?.data?.current?.aggregationsByDate?.dataByDate, bgSource);
+    // All CGM data passed into AGP should be deduplicated
+    const dataKey = bgSource === 'cbg' ? 'cbgDeduplicated' : bgSource;
+    const bgDataByDate = _.mapValues(data?.data?.current?.aggregationsByDate?.dataByDate, dataKey);
```
henry-tp, line 54: Contrary to what we tried in our meeting Clint, I found that the two earlier references to `bgSource` in this file should not be changed, because they are not using bgSource as a key to the aggregated data but the raw data. Using `cbgDeduplicated` when accessing raw data causes it to throw an error.

src/utils/StatUtil.js
```diff
@@ -296,33 +306,29 @@ export class StatUtil {
...
-        this.dataUtil.setDataAnnotations(datum);
-        result += datum.sampleInterval;
-        return result;
-      },
-      0
-    );
+    let sensorUsage = 0;
+    for (let i = 0; i < cbgData.length; i++) {
+      const datum = cbgData[i];
+      this.dataUtil.setDataAnnotations(datum);
```
henry-tp, line 315: Added this call here to sync up with develop

PR #536, opened 2025-09-09
WEB-3880 - Undefined Days Error in Basics Print View

henry-tp:
[WEB-3880]

[WEB-3880]: https://tidepool.atlassian.net/browse/WEB-3880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Blip: https://github.com/tidepool-org/blip/pull/1748

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "20.8.0"
   },
   "packageManager": "yarn@3.6.4",
-  "version": "1.51.0-web-3880-undefined-days",
+  "version": "1.51.0-web-3880-undefined-days.1",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"


src/modules/print/BasicsPrintView.js
```diff
@@ -539,9 +539,13 @@ class BasicsPrintView extends PrintView {
 
         let dayType = day.type;
 
-        if (isSiteChange) {
+        if (isSiteChangeCalendarSection) {
           if (dayType === 'inRange') {
-            dayType = data[day.date] ? SITE_CHANGE : NO_SITE_CHANGE;
+            // Only show a site change if the type of site change matches the report's config
+            // E.g. If report config'd for Cannula Fills, we don't render Cartridge Changes
+            const hasSiteChangesOfMatchingType = (data[day.date]?.subtotals?.[siteChangeSource] || 0) > 0;
```
henry-tp, line 546: I was wondering what happened in tidepool web's implementation, so I poked around. I found [this code](https://github.com/tidepool-org/tideline/blob/6e1a3c562baf6f2458168f9bc8425c6568d8c216/plugins/blip/basics/components/day/ADay.js#L37-L39) in tideline which looks very similar to what I did. So I'm reasonably confident in the approach.

PR #586, opened 2026-01-22
[WEB-3632] - Add Device Names to PDF Export

henry-tp:
[WEB-3632] 

[WEB-3632]: https://tidepool.atlassian.net/browse/WEB-3632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Blip: https://github.com/tidepool-org/blip/pull/1851


clintonium-119 (changes requested):
I have a question about the rendering approach in the basics print view that needs addressing.

Also, I'm not seeing the device list additons to the bgLog and daily print views, though they seem to be required by the Jira ticket.

src/modules/print/PrintView.js
```diff
@@ -498,18 +503,30 @@ class PrintView {
   }
 
   renderCustomTextCell(tb, data, draw, column, pos, padding, isHeader) {
```
henry-tp: Previously, in the pre-draw phase where the container sizes were measured, we were always returning `' '`. This meant that we were always measuring the container to fit the height needed for a single whitespace character. We could override this by passing a number for the height, but if left to calculate the necessary height dynamically, `voilab-pdf-table` would measure based on the whitespace character. This is fine for single-line text cells (which is all we had before) but now, with dynamic-height text cells, we need to give the PDF library the actual string so it can dynamically set the height of the container.

src/modules/print/PrintView.js
```diff
@@ -833,6 +860,86 @@ class PrintView {
...
+      label = '';
+      content = '';
+      height = 0;
+    }
+
+    return {
+      label,
+      content,
+      height,
+    };
```
henry-tp: Unfortunately this is more complicated than I wanted it to be with all of the measurements. Per the requirements, the height needed to be dynamic to accommodate large device lists. This ensures that the daily and bgLog charts can adjust their content start positions based on the list length, like so.

image

src/modules/print/PrintView.js
```diff
@@ -157,6 +160,8 @@ class PrintView {
       topEdge: this.margins.top,
     };
 
+    this.deviceNamesHeader = null; // if needed, call this.generateDeviceNamesHeader();
```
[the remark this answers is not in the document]
clintonium-119: I'm fine with either approach.

src/modules/print/PrintView.js
clintonium-119: Has any product discussion come up about whether or not we should be filtering the devices list to only display those present in the data used for a given view?

It would seem to me to be ideal to only list BGM devices on this view, in particular. 
But if you did go to the trouble of filtering, it's probably not too time-consuming to also filter  the daily and basics device lists too.

clintonium-119 (commented):
This looks good, but I'd just like confirmation that the product intent is not to show the a filtered device list for each report section, so that, for instance, the BG Log view doesn't list CGMs and pumps that did not provide bg reading for that view, and so on.

src/modules/print/PrintView.js
henry-tp: Thanks @clintonium-119 , I don't know why this eluded me or I didn't think about this.

I made a discussion thread here: https://tidepoolteam.slack.com/archives/C01ELEQUVRD/p1772666366797099

The way I see it, there are two sub-questions:

- A) Do we filter only for devices for the patient that have data in the current time frame?

- B) Do we only show BGM devices on BgLog report pages? 

For Question A, if we do want to filter only for those, my intuition is to filter the `devices` array by looking to see if it exists in `matchedDevices`. 

In your opinion, do you think that approach would work for Question A?

henry-tp: @clintonium-119 okay, so the direction from product ([slack thread](https://tidepoolteam.slack.com/archives/C01ELEQUVRD/p1772666366797099)):

Show _all_ devices (regardless of bgm/cgm) but only those with data for the current time frame.

I've made the change to match those with matchedDevices here: 6380d150aa4620fc8981e118b62b56e9dd12c27c

henry-tp:
@clintonium-119 ready for another look. Added filtering by matchedDevices

henry-tp:
Hey @clintonium-119 , this is ready for another look

clintonium-119 (dismissed):
LGTM 🔥

PR #601, opened 2026-03-09
WEB-4406 - Fix window for avg insulin calculation

henry-tp:
[WEB-4406]

Blip: https://github.com/tidepool-org/blip/pull/1889

[WEB-4406]: https://tidepool.atlassian.net/browse/WEB-4406?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/StatUtil.js
+++ b/src/utils/StatUtil.js
@@ -103,10 +103,10 @@ export class StatUtil {
 
     const combinedInsulinData = [...rawBasalData, ...bolusData, ...insulinData];
     const datumTimestamps = _.map(combinedInsulinData, datum => moment.utc(datum.time).valueOf());
-    const [start, _end] = this.endpoints;
+    const [startEndpoint] = this.endpoints;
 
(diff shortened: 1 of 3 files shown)


src/utils/StatUtil.js
```diff
@@ -103,10 +103,10 @@ export class StatUtil {
 
     const combinedInsulinData = [...rawBasalData, ...bolusData, ...insulinData];
     const datumTimestamps = _.map(combinedInsulinData, datum => moment.utc(datum.time).valueOf());
-    const firstDatumAt = _.min(datumTimestamps);
+    const [startEndpoint] = this.endpoints;
```
henry-tp, line 106: The only material change here is that we are starting the bucket formation from the startEndpoint instead of the first datum time. This way the bucket aligns with the chart time window

krystophv (approved):
LGTM 🚀

PR #602, opened 2026-03-09
[WEB-4278] Fix NaN values in horizontal bar chart data causing missing bar rendering

clintonium-119:
[WEB-4278]

When navigating to a date range before data has been fetched, the Time In Range / Readings In Range bar charts fail to render and never recover, even after valid data arrives. The numeric labels (e.g. "99%") render correctly because they look up values directly from the current chart data array by index, bypassing Victory's animation state. The bars, however, depend on Victory's animated `_y` values, which get stuck on NaN because Victory cannot interpolate from NaN to a valid number.

The root cause is `reconcileTIRDatumValues()`, which divides each datum value by `total` to compute percentages. When there's no data, `total` is `0`, producing `NaN` that overwrites the datum values. The fix adds an early return when `total` is not a positive finite number, avoiding the division by zero and preserving the original `-1` placeholder values that the Stat component already treats as "no data" (rendering empty bars and `--` labels).

Related PR: https://github.com/tidepool-org/blip/pull/1890

[WEB-4278]: https://tidepool.atlassian.net/browse/WEB-4278?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "20.8.0"
   },
   "packageManager": "yarn@3.6.4",
-  "version": "1.55.1-rc.2",
+  "version": "1.55.1-rc.3",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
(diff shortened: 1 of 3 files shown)


henry-tp (approved):
good solve

PR #610, opened 2026-04-15
[WEB-4177, WEB-4179] - Upgrades to address CVE reports

henry-tp:
Impact: Low

WEB-4179: 
`tar-fs` upgrade is the same as https://github.com/tidepool-org/blip/pull/1898 and addresses the same core concern.

WEB-4177: 
The core vuln is in the `ip.isPublic` method of the `ip` package. The `ip` package has not been patched past `2.0.1` so we cannot simply upgrade it, we must remove it

The package is pulled in storybook but the method is not called by storybook.
(https://github.com/storybookjs/storybook/pull/27529#issuecomment-2149260392). Given that storybook is a devDependency it also would not pose a risk on production. Storybook was upgraded from `7.5.0` to `7.6.20`, the next version that contains the fix. The next major versions contain significant breaking changes, so I decided not to upgrade extensively.(https://github.com/storybookjs/storybook/releases/tag/7.6.20)

The socks package has a transitive dependency to `ip`. The resolution for the socks package was bumped to `^2.7.3` to which is the first version the ip package was patched out of (https://github.com/JoshGlazebrook/socks/pull/94). The ip package was brought in by socks via multiple packages that depend on socks (chromedriver, node-gyp, etc) so it makes sense to upgrade socks.


yarn.lock
```diff
@@ -10493,13 +11686,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"ip@npm:^2.0.0":
```
henry-tp, line 10496: `ip` is removed

clintonium-119 (approved):
LGTM 💯

PR #629, opened 2026-05-08
[WEB-4562] Use the most recent upload in the datautil deviceUpload map.

clintonium-119:
[WEB-4562] 

Fixes a bug where deviceUploadMap retained the first upload seen for each deviceId during ingestion, ignoring newer uploads.

As a result, devices whose data referenced an older upload would resolve to that stale entry, or sometimes an upload that had not even been fetched, preventing the custom "Trio" label from being applied when the latest upload was the one carrying the Trio client metadata.

The map now tracks the upload from the most recent datum per deviceId.

Related PR: http://github.com/tidepool-org/blip/pull/1918

[WEB-4562]: https://tidepool.atlassian.net/browse/WEB-4562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/test/utils/DataUtil.test.js
+++ b/test/utils/DataUtil.test.js
@@ -5701,7 +5701,7 @@ describe('DataUtil', () => {
       expect(result.size).to.equal(38);
 
       expect(result.devices).to.eql([
-        { id: 'Test Page Data - 123' },
+        { bgm: false, cgm: false, oneMinCgmSampleInterval: false, id: 'Test Page Data - 123', label: 'Trio', pump: false, serialNumber: undefined },
         { id: 'AbbottFreeStyleLibre-XXX-XXXX' },
         { id: 'Dexcom-XXX-XXXX' },
         { id: 'OneTouch-XXX-XXXX' },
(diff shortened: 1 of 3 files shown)


test/utils/DataUtil.test.js
```diff
@@ -5676,7 +5701,7 @@ describe('DataUtil', () => {
       expect(result.size).to.equal(38);
 
       expect(result.devices).to.eql([
-        { id: 'Test Page Data - 123' },
+        { bgm: false, cgm: false, oneMinCgmSampleInterval: false, id: 'Test Page Data - 123', label: 'Trio', pump: false, serialNumber: undefined },
```
clintonium-119, line 5704: Adding a note as to why this changed, since it wasn't obvious to me: 

The deviceUploadMap behavior change made 'Test Page Data - 123' (the default deviceId from data/types.js:33) correctly resolve to the latest matching upload. 

The default  fixtures include 5 uploads with that deviceId (upload-1 through upload-5), and upload-5 is the Trio upload at 2018-02-05, which now wins. 

Previously the map was set on first datum seen — typically a basal/bolus with the default unmatched uploadId: 'uploadId123' — so setDevices returned just { id: 'Test Page Data - 123' }.

krystophv (approved):
LGTM :shipit:

PR #630, opened 2026-05-08
[WEB-4562] Use the most recent upload in the datautil deviceUpload map - part 2.

clintonium-119:
[WEB-4562]  A second edge case with the Trio data we have to test with meant I had to expand on the previous PR.

Some data types (e.g. basal, dosingDecision) are ingested without a deviceId, so deviceUploadMap can end up pinned to an older uploadId referenced by data that does carry deviceId. When that older upload hasn't been fetched, setDevices couldn't resolve a label — most visibly, Trio devices fell back to a bare { id } entry instead of the 'Trio' label.

setDevices now builds a deviceId → newest fetched upload lookup from uploadToDeviceIdMap and uses it as a fallback when the primary uploadsById lookup misses, so labels resolve as long as any fetched upload is associated with the deviceId.

Related PR: https://github.com/tidepool-org/blip/pull/1919

[WEB-4562]: https://tidepool.atlassian.net/browse/WEB-4562?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
     "node": "20.8.0"
   },
   "packageManager": "yarn@3.6.4",
-  "version": "1.55.2-rc.4",
+  "version": "1.55.2-rc.5",
   "description": "Tidepool data visualization for diabetes device data.",
   "keywords": [
     "data visualization"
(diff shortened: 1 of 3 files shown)


krystophv (approved):
LGTM 🚢

PR #635, opened 2026-05-12
WEB-4345 - Resolve Vanta vulnerability for `tar`

henry-tp:
[WEB-4345]
[WEB-4342]
[WEB-4385]

[WEB-4345]: https://tidepool.atlassian.net/browse/WEB-4345?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[WEB-4342]: https://tidepool.atlassian.net/browse/WEB-4342?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[WEB-4385]: https://tidepool.atlassian.net/browse/WEB-4385?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


package.json
```diff
@@ -176,6 +176,7 @@
     "lodash": "4.17.21",
     "tar-fs": "^2.1.4",
     "prismjs": "1.29.0",
-    "socks": "^2.7.3"
+    "socks": "^2.7.3",
+    "tar": "^7.5.7"
```
krystophv, line 180: My preference would be to use `yarn why tar` to investigate what it would take for our more direct dependencies to be updated to versions that meet the transitive `tar` version requirement instead of relying on an ever-growing list of `resolutions`. At a minimum, though, we should be pinning the resolutions versions to exact versions.

krystophv (changes requested):
Did we look into what the version bumps for our direct dependencies would look like to get `tar` up to spec?

PR #637, opened 2026-05-19
WEB-3632 - QA Feedback

henry-tp:
blip: https://github.com/tidepool-org/blip/pull/1927


test/helpers/mockStore.js
```diff
@@ -0,0 +1,7 @@
+export const mockStore = (state = {}) => ({
+  getState: () => state,
+  dispatch: jest.fn(),
+  subscribe: jest.fn(() => jest.fn()),
+});
+
+export default mockStore;
```
henry-tp: Viz currently does not have `redux-mock-store` installed. Given how seldom we use components in Viz (and likely will not expand upon that practice), I would not like to install `redux-mock-store`, and this minimal hand-rolled implementation should be enough.

src/modules/print/PrintView.js
```diff
@@ -711,16 +713,15 @@ class PrintView {
...
-    const isPageBreakAtTableStart = !_.isNil(tableLabel) && _.isNil(tableData);
-
-    if (isPageBreakAtTableStart) return; // prevent double header on new page
+    if (row?._isHeader) return; // eslint-disable-line no-underscore-dangle
 
     tb.addHeader();
   }
 
+  onHeaderAdd(tb, row) {
+    row._isHeader = true; // eslint-disable-line no-underscore-dangle, no-param-reassign
```
henry-tp, line 722: The previous solution didn't work (I changed the type of table row) and unfortunately I don't have much to grab onto in terms of metadata to be able detect when a row is a header. I'm not a huge fan of mutating a parameter like this but I've tested it across a variety of scenarios and it seems to be safe, Claude says it's okay as well.

clintonium-119 (changes requested):
Looks good to me.  Should probably add a small test update as per the copilot review comment for completeness, and then I'll approve

henry-tp:
Awesome, thanks @clintonium-119 

Changes: 4287719c8601efdd609815627d89c71d45e1a05f

clintonium-119 (approved):
LGTM 💯

PR #641, opened 2026-06-01
[WEB-4597] Fix edited-carb display in tooltips + daily print

clintonium-119:
[WEB-4597] 

Related PRs:
https://github.com/tidepool-org/blip/pull/1940
https://github.com/tidepool-org/tideline/pull/543

Fixes QA defects in how carb **edits** (amount, eaten-time, deletion) show in the daily bolus/carb tooltips and PDF.

## Mental model for review

A carb entry is a **`food`** datum (the thing the carb icon draws). Each bolus and the pump's reasoning are **`dosingDecision`** datums, each carrying a nested **`food`** snapshot and — once an edit has happened — an **`originalFood`** snapshot. So the carbs/time for any given moment live on *snapshots inside decisions*, not just on the food datum.

When a carb is **edited**, the platform mints a **new `food` id each time and only the final one survives**; the prior decisions stick around. So an edit chain is **one surviving `food` + N `dosingDecision`s**, and those decisions reference `food`/`bolus` ids that no longer resolve (dangling) — which is why several fixes are about *re-linking* a bolus/food to the right decision rather than trusting the stored association.

The crux: **the two devices store the snapshots inverted.**

| | `dosingDecision.food` | `dosingDecision.originalFood` |
|---|---|---|
| **twiist** | rewritten to the **final** value | the **at-bolus** snapshot (value when this bolus fired) |
| **Tidepool/DIY Loop** | the **at-decision** value (what this bolus dosed on) | the **prior** value |

Worked example — carbs entered 25 g, edited to 35 g, then 34 g, with a bolus at each:

- The bolus that dosed for 35 g: on **Loop** that's `decision.food = 35` (read `food`); on **twiist** that's `decision.originalFood = 35` while `decision.food` has been rewritten to the final 34 (read `originalFood`).
- The struck-through "initial" 25 g is always on the **earliest** decision (`originalDosingDecision`), regardless of device.

So "which carbs/eaten-time does *this* bolus or decision relate to?" depends on **device + position in the edit chain** — no single field is universally right. This PR resolves that **once, in `DataUtil.js`**, annotating each datum with the device-correct values (`carbInput`, `originalDosingDecision`, a canonical `foodTime`, and `tags.carbsEdited`/`entryTimeDiffers`/`foodTimeDiffers`); the tooltips and print view just read those annotations. **Review the data layer first — the rest follows from it.**

**Changes:**
- **Per-bolus carb amount** device-gated in `joinBolusAndDosingDecision` (Loop/DIY → `food` first; twiist → `originalFood` first) so each bolus shows the carbs *it* was dosed on.
- **Original-entry bolus** no longer skipped over a *dangling* bolus association (was rendering "only Delivered" instead of carbs + IOB).
- **Edited/deleted carbs** tagged `carbsEdited` from the earliest decision → "Total Carbs (Edited/Deleted)" + "Initial Carb Amount"; net-0 still renders.
- **Time-edit lineage** recovered via a bounded heuristic (same upload + carb net + orphaned eat-time within **30 min**) → "Time Entered/Edited" + dashed icon. *Best-effort; worst case is a wrong display timestamp, never a dose.*
- **Bolus "(eaten at …)"** now reads the meal's **canonical** eaten time (surviving food's time, shared across the lineage) so all of a meal's boluses agree — matches Figma `5827-161`.
- **`getTarget`** returns `null` (not `NaN–NaN`) with no Loop/Trio schedule; arg-order fix.

**Follow-ups (not in this PR):** deleted Loop carbs render nothing — upstream data gap (no surviving food/deletion record, unlike twiist)

[WEB-4597]: https://tidepool.atlassian.net/browse/WEB-4597?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/DataUtil.js
+++ b/src/utils/DataUtil.js
@@ -116,6 +116,8 @@ export class DataUtil {
     this.startTimer('addData');
 
     this.bolusDatumsByIdMap = this.bolusDatumsByIdMap || {};
+    this.foodDatumTimes = this.foodDatumTimes || new Set();
+    this.canonicalFoodTimeByDecisionId = this.canonicalFoodTimeByDecisionId || {};
     this.bolusToWizardIdMap = this.bolusToWizardIdMap || {};
     this.knownDeviceIds = this.knownDeviceIds || new Set();
     this.uploadToDeviceIdMap = this.uploadToDeviceIdMap || {};
@@ -331,6 +333,12 @@ export class DataUtil {
       this.bolusDatumsByIdMap[d.id] = d;
     }
 
+    // Track surviving food eat-times so joinFoodAndDosingDecision can recognize a
+    // dosing decision whose food.time was superseded by an edit (matches no surviving food).
+    if (d.type === 'food' && _.isFinite(d.time)) {
+      this.foodDatumTimes.add(d.time);
+    }
+
     if (d.type === 'pumpSettings') {
       this.pumpSettingsDatumsByIdMap[d.id] = d;
     }
@@ -396,9 +404,17 @@ export class DataUtil {
         const proximateDosingDecisions = _.filter(
           _.mapValues(this.bolusDosingDecisionDatumsByIdMap),
           ({ time, associations }) => {
-            // If there is a definitive association, we skip this decision, as it would have been
-            // associated with the bolus already in the code above if the id matched
-            if (_.some(associations, { reason: 'bolus' })) return false;
+            // Skip a decision only if it is already claimed by a DIFFERENT, real bolus —
+            // that bolus would have matched it via the definitive association above. We
+            // must NOT skip on a dangling `bolus` association whose id resolves to no
+            // normalized bolus datum (e.g. twiist references an original-entry bolus by an
+            // id that doesn't survive normalization); otherwise the carb's original-entry
+            // bolus can never match its decision and renders with no carbs/IOB.
+            const claimedByOtherBolus = _.some(
+              associations,
+              a => a.reason === 'bolus' && a.id !== d.id && !!this.bolusDatumsByIdMap[a.id]
+            );
+            if (claimedByOtherBolus) return false;
 
             const timeOffset = Math.abs(time - d.time);
             return timeOffset <= timeThreshold;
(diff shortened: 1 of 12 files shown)


src/utils/bolus.js
```diff
@@ -378,7 +378,13 @@ export function isCorrection(insulinEvent) {
   const recommended = insulinEvent.dosingDecision
     ? {
       correction: _.get(insulinEvent, 'dosingDecision.recommendedBolus.amount'),
-      carb: _.get(insulinEvent, 'dosingDecision.food.nutrition.carbohydrate.net', 0),
+      // Prefer originalFood (immutable pre-edit snapshot) over food, which may be a
+      // post-edit value on intermediate DDs in an edit chain.
```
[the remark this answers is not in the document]
clintonium-119: Was a good catch.  Fixed.

src/utils/DataUtil.js
```diff
@@ -425,8 +441,20 @@ export class DataUtil {
...
-        d.carbInput = d.dosingDecision.food?.nutrition?.carbohydrate?.net;
+        // Carbs the bolus was computed on — the field differs by device:
+        //   • twiist rewrites every prior decision's `food` to the FINAL value and stores
+        //     the at-bolus snapshot in `originalFood`, so `originalFood` is the value at
+        //     this bolus's decision (prefer it).
+        //   • Tidepool / DIY Loop instead leave each decision's `food` at the value that
+        //     decision (and its bolus) was computed on, with `originalFood` holding the
+        //     PRIOR value — so prefer `food`, else an interim-edit bolus surfaces the
+        //     pre-edit count (e.g. 25g on a bolus actually dosed for 35g).
+        const carbsFromFoodFirst = isTidepoolLoop(d) || isDIYLoop(d);
```
krystophv, line 452: Do Trio devices also behave in the TidepoolLoop/DIYLoop manner?

clintonium-119: I asked Bill, and we don't actually know.  He's asking that we leave it out of scope for this release.

krystophv (approved):
This was clearly kinda rough. The opposing behaviors from two otherwise very similar data sources is pretty annoying. Great work getting it ironed out.

LGTM :shipit:

PR #651, opened 2026-06-25
[WEB-55] Daily view site change events

clintonium-119:
[WEB-55] 

Related PRs:
https://github.com/tidepool-org/blip/pull/1965
https://github.com/tidepool-org/tideline/pull/553

[WEB-55]: https://tidepool.atlassian.net/browse/WEB-55?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/components/daily/eventtooltip/EventTooltip.js
+++ b/src/components/daily/eventtooltip/EventTooltip.js
@@ -7,7 +7,7 @@ import Tooltip from '../../common/tooltips/Tooltip';
 import colors from '../../../colors';
 import detailedEventStyles from './DetailedEventTooltip.css';
 import standardEventStyles from './StandardEventTooltip.css';
-import { EVENT_HEALTH, EVENT_NOTES, EVENT_PHYSICAL_ACTIVITY, EVENT_PUMP_SHUTDOWN, MS_IN_MIN } from '../../../utils/constants';
+import { EVENT_HEALTH, EVENT_NOTES, EVENT_PHYSICAL_ACTIVITY, EVENT_PUMP_SHUTDOWN, MS_IN_DAY, MS_IN_MIN } from '../../../utils/constants';
 import { formatClocktimeFromMsPer24, formatDuration, getMsPer24 } from '../../../utils/datetime';
 
 import tandemShutDownImage from './images/tandemShutDownImage.png';
@@ -45,6 +45,21 @@ const renderStandardEvent = (content = {}) => (
 const getEventContent = (event, timePrefs) => {
   const msPer24 = getMsPer24(event?.normalTime, timePrefs?.timezoneName);
   const time = formatClocktimeFromMsPer24(msPer24);
+
+  if (event?.tags?.siteChange) {
+    const daysSinceText = event.daysSince != null
+      ? formatDuration(event.daysSince * MS_IN_DAY, { condensed: true })
+      : null;
+
+    return {
+      time,
+      title: event.displayLabel
+        ? t('Site Change: {{displayLabel}}', { displayLabel: event.displayLabel })
+        : t('Site Change'),
+      value: daysSinceText,
+      renderer: renderStandardEvent,
+    };
+  }
 
   switch (event?.tags?.event) {
     case EVENT_PUMP_SHUTDOWN:
@@ -151,7 +166,7 @@ EventTooltip.propTypes = {
   showDividers: PropTypes.bool,
   event: PropTypes.shape({
     tags: PropTypes.shape({
-      event: PropTypes.string.isRequired,
+      event: PropTypes.string,
     }).isRequired,
     normalTime: PropTypes.number.isRequired,
   }).isRequired,
--- a/src/index.js
+++ b/src/index.js
@@ -95,6 +95,7 @@ import {
   findBasicsStart,
   defineBasicsAggregations,
   processBasicsAggregations,
+  getSiteChangeSource,
 } from './utils/basics/data';
 
 const i18next = require('i18next');
@@ -186,6 +187,7 @@ const utils = {
   aggregation: {
     defineBasicsAggregations,
     processBasicsAggregations,
+    getSiteChangeSource,
   },
   text: {
     TextUtil,
--- a/src/modules/print/DailyPrintView.js
+++ b/src/modules/print/DailyPrintView.js
@@ -61,6 +61,7 @@ import {
 import {
   ALARM,
   AUTOMATED_DELIVERY,
+  DIY_LOOP,
   EVENT_HEALTH,
   EVENT_NOTES,
   EVENT_PHYSICAL_ACTIVITY,
@@ -69,8 +70,15 @@ import {
   PHYSICAL_ACTIVITY,
   PREPRANDIAL,
   SCHEDULED_DELIVERY,
+  SITE_CHANGE_CANNULA,
+  SITE_CHANGE_RESERVOIR,
+  SITE_CHANGE_TUBING,
   SLEEP,
+  TIDEPOOL_LOOP,
+  TWIIST_LOOP,
 } from '../../utils/constants';
+
+import { getSiteChangeSource } from '../../utils/basics/data';
 
 import {
   processBasalRange,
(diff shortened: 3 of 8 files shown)


krystophv:
@coderabbitai review

src/components/daily/eventtooltip/EventTooltip.js
```diff
@@ -46,6 +46,21 @@ const getEventContent = (event, timePrefs) => {
...
   const time = formatClocktimeFromMsPer24(msPer24);
 
+  if (event?.tags?.siteChange) {
+    const daysSinceText = event.daysSince != null
+      ? formatDuration(event.daysSince * MS_IN_DAY, { condensed: true })
+      : null;
+
+    return {
+      time,
+      title: event.displayLabel
```
krystophv, line 59: Nothing on the Daily path sets `displayLabel` or `daysSince`. tideline sends the bare datum from `js/plot/sitechange.js:162-170`, and blip only adds positioning. Grepping both fields across viz, tideline and blip turns up only the two reads here.

Tagged a real `prime`/`primeTarget: 'cannula'` datum the way `DataUtil.js:839-851` does and rendered it:

```
as shipped:        "1:45 pmSite Change"
with both set:     "1:45 pmSite Change: Cannula Fill3d 6h"
```

So `tags.siteChange` is true and the branch is reached, but both fields are `undefined` and the browser shows two lines where the design asks for three.

The natural fix looks like attaching both to the hover payload in tideline, with the label from `getSiteChangeSourceLabel` and `daysSince` from `DataUtil.getPreviousSiteChangeDatums`.

src/components/daily/eventtooltip/EventTooltip.js
```diff
@@ -46,6 +46,21 @@ const getEventContent = (event, timePrefs) => {
   const msPer24 = getMsPer24(event?.normalTime, timePrefs?.timezoneName);
   const time = formatClocktimeFromMsPer24(msPer24);
 
+  if (event?.tags?.siteChange) {
+    const daysSinceText = event.daysSince != null
+      ? formatDuration(event.daysSince * MS_IN_DAY, { condensed: true })
```
krystophv: `condensed` makes `formatDuration` read the day _component_ of the duration rather than the total, so it wraps past a month. Swept it:

```
 3.5 -> "3d 12h"
  29 -> "29d"
  30 -> "30d"
  31 -> "0m"
  45 -> "14d"
  61 -> "0m"
  90 -> "29d"
 365 -> "30d"
```

`datetime.js:234` takes `momentDuration.days()`, which is the remainder after months, and nothing reads `.months()`. At exactly 31 days that leaves `0d 0h 0m`, so it falls into the less-than-a-minute branch on line 253 and prints `0m`.

Invisible today only because `daysSince` is never set, so whoever populates that field inherits this unless they change it too. Basics renders whole days, which is also what the design asks for.

In practice, a site change _shouldn't_ get near that boundary, but this could accidentally mask pathologic/improper data if the calculated duration were to extend that long.

src/components/daily/eventtooltip/EventTooltip.js
```diff
@@ -151,7 +166,7 @@ EventTooltip.propTypes = {
   showDividers: PropTypes.bool,
   event: PropTypes.shape({
     tags: PropTypes.shape({
-      event: PropTypes.string.isRequired,
+      event: PropTypes.string,
```
krystophv, line 172: nit: the shape still only declares `tags.event` and `normalTime`, but the branch on line 49 reads `tags.siteChange`, `displayLabel` and `daysSince`. Since relaxing `event` to optional here was needed precisely because site changes don't carry it, seems worth declaring the three fields that replace it.

test/components/daily/EventTooltip.test.js
```diff
@@ -119,13 +98,41 @@ describe('EventTooltip', () => {
...
+      expect(screen.getByText('Had two drinks')).to.exist;
+    });
+  });
+
+  context('site change event', () => {
+    const event = {
+      tags: { siteChange: true },
+      normalTime: new Date('2023-01-01T12:00:00Z').valueOf(),
+      displayLabel: 'Cannula Fill',
+      daysSince: 3.5,
```
krystophv: This fixture sets `displayLabel` and `daysSince` by hand, which I think is why the suite stays green over a datum shape the pipeline never produces. Ran all 105 tests in the three affected files and they pass, while the rendered output is the fallback on line 133.

Worth rebuilding from a real tagged deviceEvent so it would have caught this. The `'3d 12h'` assertion below also bakes in the format from the `formatDuration` comment, so it'll need updating either way.

src/modules/print/DailyPrintView.js
```diff
@@ -90,6 +99,50 @@ const eventImages = {
...
 };
 
+// Site-change sprites. Loop uses the loop-tubing variant, twiist the
+// twiist-cassette variant; manufacturer variants keyed `<manufacturer>_<subtype>`.
+const siteChangeImages = {
+  [SITE_CHANGE_CANNULA]: 'images/sitechange-cannula.png',
+  [SITE_CHANGE_RESERVOIR]: 'images/sitechange-reservoir.png',
+  [SITE_CHANGE_TUBING]: 'images/sitechange-tubing.png',
+  [`${_.lowerCase(TIDEPOOL_LOOP)}_${SITE_CHANGE_TUBING}`]: 'images/sitechange-loop-tubing.png',
+  [`${_.lowerCase(DIY_LOOP)}_${SITE_CHANGE_TUBING}`]: 'images/sitechange-loop-tubing.png',
```
krystophv: nit, and no action needed: Trio isn't in this map, so a Trio user gets the generic `tubing` icon. `getSiteChangeSource` maps it to `tubingPrime` (`basics/data.js:250`), but none of the three loop-variant lookups include it (this map, `tideline/js/plot/sitechange.js:11`, and Basics' `Change.js:53`).

Generic is likely the right outcome for now.

src/modules/print/DailyPrintView.js
```diff
@@ -90,6 +99,50 @@ const eventImages = {
...
+  if (_.get(d, 'tags.reservoirChange')) return SITE_CHANGE_RESERVOIR;
+  if (_.get(d, 'tags.cannulaPrime')) return SITE_CHANGE_CANNULA;
+  if (_.get(d, 'tags.tubingPrime')) return SITE_CHANGE_TUBING;
+  return null;
+}
+
+// Resolve the site-change image from subtype + manufacturer, matching getSiteChangeSource's
+// `_.lowerCase` normalization. Loop/twiist fall back to the base icon when no variant exists.
+function getSiteChangeImage(subType, manufacturer) {
+  return siteChangeImages[`${_.lowerCase(manufacturer)}_${subType}`] || siteChangeImages[subType];
```
krystophv: nit: this key format repeats the `_.lowerCase` normalization in `getSiteChangeSource`, and the comment above documents the coupling rather than removing it. Might be an opportunity for a shared `manufacturerKey()` to make it explicit.

src/modules/print/DailyPrintView.js
```diff
@@ -90,6 +99,50 @@ const eventImages = {
...
+  [`${_.lowerCase(TIDEPOOL_LOOP)}_${SITE_CHANGE_TUBING}`]: 'images/sitechange-loop-tubing.png',
+  [`${_.lowerCase(DIY_LOOP)}_${SITE_CHANGE_TUBING}`]: 'images/sitechange-loop-tubing.png',
+  [`${_.lowerCase(TWIIST_LOOP)}_${SITE_CHANGE_RESERVOIR}`]: 'images/sitechange-twiist-cassette.png',
+};
+
+// Resolve a deviceEvent to a SITE_CHANGE_* subtype. Daily data records site
+// changes as subType `reservoirChange` or `prime` (+ `primeTarget`); also
+// tolerate the already-keyed subTypes and boolean `tags`.
+function getSiteChangeSubType(d) {
+  if (d.subType === SITE_CHANGE_RESERVOIR) return SITE_CHANGE_RESERVOIR;
```
krystophv, line 117: nit: this is verbatim `tideline/js/plot/sitechange.js:21-30`, and Basics does the same classification a third way in `Change.js:48-55`.

The cascade itself seems right. It reconciles three encodings of one concept into an enum, which isn't fixable locally. It's the three copies that'll probably drift.

src/modules/print/DailyPrintView.js
```diff
@@ -90,6 +99,50 @@ const eventImages = {
...
+// `_.lowerCase` normalization. Loop/twiist fall back to the base icon when no variant exists.
+function getSiteChangeImage(subType, manufacturer) {
+  return siteChangeImages[`${_.lowerCase(manufacturer)}_${subType}`] || siteChangeImages[subType];
+}
+
+const SITE_CHANGE_DEDUP_WINDOW_MS = 5 * MS_IN_MIN;
+
+// Site changes within 5 minutes of one another count as the same change; keep only
+// the first (sorted by time), so at most one icon prints per 5-minute window.
+function dedupeSiteChangesWithinWindow(siteChanges) {
```
krystophv, line 136: Same algorithm as `tideline/js/plot/sitechange.js:132-144`, just a different dialect. I ran both over nine inputs (boundary cases, bursts, unsorted, duplicate timestamps) and they agree every time, so the window and the last-kept anchoring really are identical, not just similar.

Which is the problem: two homes that have to change together, and sharing the code isn't really on offer. Neither library can import the other, and blip can't bridge it here since the PDF runs in a worker against `dist/print` and never holds the deviceEvent data.

So maybe pin the behaviour instead: the same fixture of boundary cases in `test/fixtures/` here and in tideline, each asserting its own implementation, with a header naming the other path. Turns a silent divergence into a failing test, which is most of the value.

src/modules/print/DailyPrintView.js
```diff
@@ -1261,6 +1327,28 @@ class DailyPrintView extends PrintView {
       });
     });
 
+    // Site changes matching the selected subtype, alongside the other events.
+    const siteChangeImage = this.siteChangeSource
+      && getSiteChangeImage(this.siteChangeSource, this.manufacturer);
+
+    if (siteChangeImage) {
+      const siteChanges = dedupeSiteChangesWithinWindow(
```
krystophv, line 1333: `this.siteChangeSource` is resolved once in the constructor from `this.manufacturer`, so this filter uses the current pump's subtype for every day in the document. A patient who switched pumps loses the old device's site changes on every page. Same as the Daily view, and same as Basics, so the three agree and this looks like the intended reading.

Separately, and this one is a real difference: `siteChangeImage` on 1331 is computed once from `this.manufacturer` and reused inside the loop, while Daily resolves the icon per datum from `d.source`. On a day whose data came from more than one upload the PDF prints one icon where Daily would draw two. Ties into the same question on tideline#553.

src/index.js
```diff
@@ -95,6 +95,7 @@ import {
   findBasicsStart,
   defineBasicsAggregations,
   processBasicsAggregations,
+  getSiteChangeSource,
```
krystophv, line 98: `getSiteChangeSourceLabel` isn't exported from here, so blip can't reach it, and the `displayLabel` fix needs it. Looks like it wants the same two lines added for `getSiteChangeSource`, here and in `utils.aggregation` below.

krystophv (changes requested):
The site-change tooltip renders, but nothing ever sets `displayLabel` or `daysSince`, so it shows only the time and the bare words "Site Change". There's a second bug hiding behind that one, in `formatDuration`. They mask each other, so fixing either alone ships something visibly wrong.

Everything else is minor.

src/modules/print/DailyPrintView.js
clintonium-119: Confirmed, keeping the current-pump subtype filter to match Basics and the Done Criteria ("Selection applies to full history", "No Basics selection = no icons"). Noted on WEB-55 that the 2017 answer is superseded.

PR #657, opened 2026-07-22
WEB-4556 add glycemic ranges util exports

krystophv:
for [WEB-4556] related export PR: https://github.com/tidepool-org/export/pull/91
This pull request refactors how glycemic range constants and utilities are organized and exposed, making them available in a dependency-free module suitable for both browser and Node.js environments. It also introduces a new utility function for resolving blood glucose (BG) bounds for a patient's glycemic range, ensures robust property checks, and adds comprehensive tests. The package version is incremented to reflect these changes.

**Glycemic Range Constants Refactor and Export:**

* Extracted all glycemic range-related constants from `src/utils/constants.js` into a new, dependency-free module `src/utils/glycemicRangeConstants.js`, and updated all imports to use this new module. This allows these constants to be bundled for Node.js consumers (e.g., data-export-service) without browser-specific dependencies. [[1]](diffhunk://#diff-bd2e652035adaf33b6125f14a823cf76feae0c1c1a592e9a6c14a14951a52a28L48-R56) [[2]](diffhunk://#diff-f479ded1de7e06a2e12d991e0acac18a7651b78e5685696fbe57d85fa65c648bR1-R98) [[3]](diffhunk://#diff-40a34648d57b38407237f9a0d64b27bd001039d615e87441757238b52cc729e5L1-R25)

**Utility Function Enhancements:**

* Added a new function `getBgBoundsForGlycemicRanges` to `src/utils/glycemicRanges.js`, which robustly resolves the correct BG bounds for a patient, with defensive checks against prototype pollution and invalid units. This function is now exported and available for both browser and Node.js usage. [[1]](diffhunk://#diff-40a34648d57b38407237f9a0d64b27bd001039d615e87441757238b52cc729e5R48-R74) [[2]](diffhunk://#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556L84-R84) [[3]](diffhunk://#diff-bfe9874d239014961b1ae4e89875a6155667db834a410aaaa2ebe3cf89820556R199-R200)

**Testing Improvements:**

* Added comprehensive unit tests for both `getGlycemicRangesPreset` and the new `getBgBoundsForGlycemicRanges` in `test/utils/glycemicRanges.test.js`, covering edge cases like missing data, invalid units, and prototype pollution.

**Build and Packaging:**

* Updated `webpack.config.js` to expose `glycemicRanges` as a standalone bundle entry, enabling external Node.js consumers to import glycemic range utilities directly.
* Bumped the package version to `1.59.0-web-4556-glycemic-ranges.1` to reflect these functional and structural changes.

[WEB-4556]: https://tidepool.atlassian.net/browse/WEB-4556?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/utils/glycemicRangeConstants.js
+++ b/src/utils/glycemicRangeConstants.js
@@ -21,8 +21,8 @@
 // Do not import i18next, lodash, or anything else from here — keeping this
 // module side-effect-free is what lets it load in plain Node.
 
-const MGDL_UNITS = 'mg/dL';
-const MMOLL_UNITS = 'mmol/L';
+export const MGDL_UNITS = 'mg/dL';
+export const MMOLL_UNITS = 'mmol/L';
 
 export const GLYCEMIC_RANGES_PRESET = {
   ADA_STANDARD: 'adaStandard',
--- a/src/utils/glycemicRangeConstants.js
+++ b/src/utils/glycemicRangeConstants.js
@@ -1,0 +1,98 @@
+/*
+ * == BSD2 LICENSE ==
+ * Copyright (c) 2026, Tidepool Project
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the associated License, which is identical to the BSD 2-Clause
+ * License as published by the Open Source Initiative at opensource.org.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+ * FOR A PARTICULAR PURPOSE. See the License for more details.
+ *
+ * You should have received a copy of the License along with this program; if
+ * not, you can obtain one from Tidepool Project at tidepool.org.
+ * == BSD2 LICENSE ==
+ */
+
+// Glycemic-range constants extracted from utils/constants.js into their own
+// dependency-free module so they can be exposed via a Node-friendly bundle
+// (dist/glycemicRanges.js) for non-browser consumers (e.g. data-export-service).
+// Do not import i18next, lodash, or anything else from here — keeping this
+// module side-effect-free is what lets it load in plain Node.
+
+const MGDL_UNITS = 'mg/dL';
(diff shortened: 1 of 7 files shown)


src/utils/glycemicRangeConstants.js
```diff
@@ -0,0 +1,98 @@
...
+ */
+
+// Glycemic-range constants extracted from utils/constants.js into their own
+// dependency-free module so they can be exposed via a Node-friendly bundle
+// (dist/glycemicRanges.js) for non-browser consumers (e.g. data-export-service).
+// Do not import i18next, lodash, or anything else from here — keeping this
+// module side-effect-free is what lets it load in plain Node.
+
+export const MGDL_UNITS = 'mg/dL';
+export const MMOLL_UNITS = 'mmol/L';
```
henry-tp, line 25: Oh man. Okay, this breaking my brain a little bit. 

I did not realize that the `MGDL_UNITS` and `MMOLL_UNITS` constants in `constants.js` were run through the `i18n.t()` fn. I think during my development process, my brain probably ignored that. I really don't like this because it's mixing together the rendered form which could be different in each language.

It feels like if everything is defined the same way (using the translated string) it should be fine? But mixing the translated vs untranslated forms might be dangerous.

Sorry, I'm just thinking out loud here. I'm going to do some exploration and play around. I'm actually kind of worried about this - not from your changes in this PR here but before it

henry-tp: Okay. After doing a bunch of digging .. I guess it seems to be okay for now, as long as we consistently don't translate these strings from their original form for any new language.

However, here, can we switch to using the definitions from `constants.js` for consistency?

krystophv: I think we needed to avoid importing i18next in the node context, which constants.js does.

henry-tp (approved):

